PR #11, opened 2024-11-19
a document about the Lattice Standard

egstern:
This is a meta document about the Lattice Standard giving my view of what should go into the standard and what should not go into it.

### DIFF
--- a/About.md
+++ b/About.md
@@ -1,0 +1,63 @@
+# About the Lattice Standard
+
+## What the Lattice Standard is
+
+1. The Lattice Standard is an *abstract* description of a lattice for the simulated transport of charged particles.
+
+1. A Lattice is an ordered list of regions in space distinguished by the presense of (possibly time-varying) specially described electromagnetic fields,
+materials, apertures and other possible engineered structures. For simplicity, we will denote these regions as _lattice_elements_ although a region
+may be distinguished by the absence of material (drift space.)
+
+1. A lattice will have the notion of a _reference_trajectory_ which is the defined path of the particle which the lattice is intended to transport.
+
+1. The positioning of lattice elements will be described by their order in the list of elements and their $s$ coordinate, the distance along the path of
+propagated reference particle. The definition and use of coordinates transverse to the direction of motion is not in this document  and will be
+the subject of the definition of the standard.
+
+1. The Lattice Standard is intended to be extensible and general enough to accommodate elements that are used in current particle transport machines.
+
+1. The way in which elements are described is not part of this document and will make up the main part of the work of the standard.
+
+1. The Lattice Standard defines a language and a document composed of words and symbols made of UTF-8 characters that is human readable.
+
+1. The language defined by the Lattice Standard should conform to a formal grammar in the the common meaning in computer science and be unambiguously parsable by machines.
+
+1. Any particular simulation code may implement transport of particles through a particular element type as described in the Lattice Standard. In that case, that element type will be said to be implemented by that particular code.
+
+1. Any particular simulation code may not implement transport of particles through a particular element type as described in the Lattice Standard. In that case, that element type will be said to be not implemented by that particular code.
+
+1. Any code may make its own choice about what parts and what elements in the Lattice Standard it chooses to implement.
+
+### Ingredients for the Lattice Standard
+
+1. It would be nice for the Lattice Standard language to have the ability to define subsections of a machine (arcs) which substitutable arguments that can be duplicated and placed within a larger structure.
+
+1. It would be nice for the Lattice Standard language to allow mechanisms to manipulate and transform defined arcs.
+
+1. It would be nice for the Lattice Standard language to allow lattice element properties to be specified with substituble parameters
+which may also be mathematical expressions.
+
+## What the Lattice Standard is not
+
+1. The Lattice Standard does not say anything about its implementation in any computer language.
+
+1. The Lattice Standard does not say anything about the algorithms or physics that goes into the implmentation of particle transport simulations.
+
+1. Any particular lattice element may include in its description hints about the way in which it is intended to be implemented. The standard does not require their usage in any code in any particular way.
+
+## Lattice Readers
+
+1. The Lattice Standard organization encourages the development of software in various languages that implement tools
+for parsing Lattice Standard documents and making the information available to particle transport codes. These tools are separate from the Lattice Standard.
+
+1. At a minimum, a Reader will parse a Lattice Standard document and provide and interface for a code to request the list of elements with their characteristics as defined in the document in the
+form which is appropriate for the language in which is implemented. The standard does not prescribe the way that this is done.
+
+1. If a reader encounters a document purporting to be a Lattice Standard document that it is unable to parse, it may choose its own action among possibilities such as
+    1. Fail
+    2. Parse the elements that it can
+    3. Attempt corrective actions
+
+1. Each reader will define its own interface in its own documentation.
+
+


About.md
```diff
@@ -0,0 +1,63 @@
+# About the Lattice Standard
+
+## What the Lattice Standard is
+
+1. The Lattice Standard is an *abstract* description of a lattice for the simulated transport of charged particles.
+
+1. A Lattice is an ordered list of regions in space distinguished by the presense of (possibly time-varying) specially described electromagnetic fields,
+materials, apertures and other possible engineered structures. For simplicity, we will denote these regions as _lattice_elements_ although a region
```
DavidSagan, line 8: Instead of "For simplicity, we will denote these regions as _lattice_elements_ " better I think is "These regions will be called _lattice elements_

About.md
```diff
@@ -0,0 +1,63 @@
...
+may be distinguished by the absence of material (drift space.)
+
+1. A lattice will have the notion of a _reference_trajectory_ which is the defined path of the particle which the lattice is intended to transport.
```
DavidSagan, line 11: This is a bit confusing. Especially since the particles may never follow the reference trajectory (EG: in a wiggler the reference trajectory is straight but the actual particle path is oscillatory). I think better is:

A lattice will have the notion of a _reference_trajectory_, also called the zero orbit, about which the curvilinear coordinate system of the lattice is defined. This curvilinear coordinate system, called _machine coordinates_ in the standard, establishes a coordinate system which may be used to describe particle trajectories and lattice element alignment.

About.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. The positioning of lattice elements will be described by their order in the list of elements and their $s$ coordinate, the distance along the path of
+propagated reference particle. The definition and use of coordinates transverse to the direction of motion is not in this document  and will be
+the subject of the definition of the standard.
```
DavidSagan, line 15: I am confused by "The definition... of the standard.". The transverse coordinates are needed to define how lattice elements are oriented with respect to each other. I would leave this sentence out.

About.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. Any particular simulation code may not implement transport of particles through a particular element type as described in the Lattice Standard. In that case, that element type will be said to be not implemented by that particular code.
+
+1. Any code may make its own choice about what parts and what elements in the Lattice Standard it chooses to implement.
+
+### Ingredients for the Lattice Standard
+
+1. It would be nice for the Lattice Standard language to have the ability to define subsections of a machine (arcs) which substitutable arguments that can be duplicated and placed within a larger structure.
+
+1. It would be nice for the Lattice Standard language to allow mechanisms to manipulate and transform defined arcs.
```
DavidSagan, line 35: This gets into a discussion of portability versus usability. Portability is maximized by minimizing manipulations that are needed to instantiate a lattice within a program. I think we can eventually include such manipulations but I would  propose we initially aim for portability.

About.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. It would be nice for the Lattice Standard language to allow lattice element properties to be specified with substituble parameters
+which may also be mathematical expressions.
+
+## What the Lattice Standard is not
+
+1. The Lattice Standard does not say anything about its implementation in any computer language.
```
DavidSagan, line 42: Yes the Standard itself will be language agnostic. To make the Standard useful, we will need to create translating software that will be housed in the repo (or repos) with the Standard.

About.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. The Lattice Standard does not say anything about the algorithms or physics that goes into the implmentation of particle transport simulations.
```
DavidSagan, line 44: Yes this is an important point. The Standard will need to rigorously define everything. For example, what is the field associated with a multipole.

About.md
```diff
@@ -0,0 +1,63 @@
...
+
+1. Any particular lattice element may include in its description hints about the way in which it is intended to be implemented. The standard does not require their usage in any code in any particular way.
```
DavidSagan, line 46: I'm confused by what is intended here.

About.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Lattice Readers
+
+1. The Lattice Standard organization encourages the development of software in various languages that implement tools
+for parsing Lattice Standard documents and making the information available to particle transport codes. These tools are separate from the Lattice Standard.
```
DavidSagan, line 51: To make the Standard usable, Readers/Writers should be part of the project. Just as Readers/Writers are part of OpenPMD.

DavidSagan (commented):
Hi Eric: Very nice and much thanks for this. I have left some comments for your consideration.

About.md
```diff
@@ -0,0 +1,63 @@
...
+propagated reference particle. The definition and use of coordinates transverse to the direction of motion is not in this document  and will be
+the subject of the definition of the standard.
+
+1. The Lattice Standard is intended to be extensible and general enough to accommodate elements that are used in current particle transport machines.
+
+1. The way in which elements are described is not part of this document and will make up the main part of the work of the standard.
+
+1. The Lattice Standard defines a language and a document composed of words and symbols made of UTF-8 characters that is human readable.
+
+1. The language defined by the Lattice Standard should conform to a formal grammar in the the common meaning in computer science and be unambiguously parsable by machines.
```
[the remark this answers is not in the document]
egstern: Isn't a schema a kind of grammar as well? I'm not familiar with the details of what makes something a schema and whether it is too constraining. I would like to see some examples of how that would work.

About.md
egstern: Yes, I was unsure of how to deal with things like undulators and even what MAD calls RBENDS. We definitely need some language that can deal with it but it wasn't obvious to me without maybe a figure what your terminology refers to and how you would use it to place elements at positions.

About.md
egstern: Sounds good to me.

About.md
egstern: OK, but I don't want to specify an interface that a reader/writer has to follow, especially because the natural idiom for doing things can vary by language, not to mention that the representation of all the data objects is going to be different in different languages.

About.md
DavidSagan: Yes I plan on submitting a PR that will go into this in detail.

About.md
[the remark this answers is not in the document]
DavidSagan: Agreed. I believe things will become much clearer when we start fleshing out the Standard.

About.md
DavidSagan: Yes. This is not preventing someone from implementing their own. We just don't want people to have to reimplement the wheel so to speak.

About.md
rdemaria: Point 2 and 4 seem too specific to me in this context, as they imply a specific choice of the relation between lattice, elements and positioning. I would state for instance that “the standard will define what is a lattice and how it can be described in terms of primitive components".

For instance, you can call a "lattice element" a coordinate transformation, observation point, a matrix element that are not regions of space, but which you could order. Conversely, you may call elements, specifications of certain aspects of a region of space (fields, aperture, etc,), without an explicit positioning and the lattice a unordered collection of the positioning of existing elements relative to a reference trajectory.

DavidSagan:
This document, as Eric says, is a view on how to develop the Standard. As such, as the Standard is developed, I'm sure our conception of what the standard is, and is not, will evolve. So I don't take this as a mandate but rather something to help guide our thoughts. Given this, I'm thinking that it would be more useful to move this to the Conversation area where there can be a conversation that is not limited in time.

DavidSagan:
Since Eric has moved this to the Discussions area, I am closing the PR.

PR #19, opened 2024-12-03
[WIP] Add docs prototype (MyST)

EZoni:
Prototype of documentation based on MyST.

Follow instructions in `section_documentation.md` to build and display the documentation locally.

Options for deployment (static website vs application) can be investigated separately, starting from the MyST documentation at https://mystmd.org/guide/deployment.

As part of this, I have fixed the syntax in some of the existing markdown files.

Here's a screenshot of the math example section:
Screenshot from 2024-12-03 15-11-04


README.md
EZoni, line 1: Given that we currently do not provide a file named `index.md`, this README file is the root file used as home page when building the documentation with MyST. See https://mystmd.org/guide/table-of-contents#root-page.

DavidSagan (approved):

cemitch99 (approved):
Looks good to me.

PR_Instructions.md
EZoni, line 1: Note that I simply renamed this file as `CONTRIBUTING.md`, which I think is a bit more standard.

PR #32, opened 2025-01-14
Coordinate systems standard

DavidSagan:
Description of the coordinate systems to be used with the lattice standard.


MichaelEhrlichman:
The following sentence would be more clear of "global" were replaced with "floor".  "Global coordinates" appears only one time in the document.
_Fiducial elements only have a single coordinate frame that is tied to global coordinates and construction of the branch coordinate system starts at this coordinate system._

DavidSagan:
@MichaelEhrlichman Done.

PR #34, opened 2025-01-20
Start beamline standard devel.

DavidSagan:
Start devel of element parameters standard.


source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+y_limit
+aperture_shape
+aperture_at
+aperture_shifts_with_body
+wall2d
+material
+thickness
+```
+
+### aperture_at
```
cemitch99, line 69: I think the name `aperture_at` can be confusing; my first instinct is to assume that `at` is an abbreviation.  I would propose something straightforward and explicit, like `aperture_location`.

DavidSagan: Or how about just `location`?

source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+
+For an element to inherit all parameter groups from another element, just inherit the element itself:
+```{code} yaml
+element:
+  name: q3
+  inherit: q2
+```
+
+## Aperture Parameter Group
+
```
cemitch99, line 56: As a general comment:  some of these names for components begin with the prefix `aperture_` , while others do not.  For consistency, why not use this prefix for _all_ of the components, or for _none_?  Either option would work, but we could follow the same convention for other parameter groups as well.

source/element_parameters.md
DavidSagan: And `aperture_shape` can be shortened to `shape`.

source/element_parameters.md
DavidSagan: Yes whatever looks best. I would use `shape` and `location` but am still partial to `aperture_shifts_with_body` (but not strongly so).

source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+  7) End at downstream end of element.
+```
+
+### material
+
+The `material` parameter sets the material of the aperture. 
+Using chemical formulas like `Cu` and `Fe` are the most portable.
+
+### wall2d
+
```
cemitch99, line 157: There could be a more descriptive name also for `wall2d`, such as `user_defined`, `aperture_user_defined`, or something similar.  Some name that reflects the flexibility to construct a very general aperture boundary.

DavidSagan: One possibility is `vertex`. `wall2d` is in reference to the fact that this description can be extended to `wall3d` which defines a 3D wall using an array of `wall2d` at different s-positions.

source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+```
+For a `RECTANGULAR` aperture the corresponding four conditions are:
+```{code}
+  1) x < x_limit[1]
+  2) x > x_limit[2]
+  3) y < y_limit[1]
+  4) y > y_limit[2]
+```
+
+Default values for the limits are `[-Inf, Inf]` for both `x_limit` and `y_limit`.
```
cemitch99, line 121: Recommend adding an explicit specification of units here.  (Presumably meters.)

source/element_parameters.md
DavidSagan: Also there needs to be a distinction between `wall2d` and a custom program dependent defined aperture that is not described by the standard

source/element_parameters.md
DavidSagan: See the new introduction for a discussion on units. One thing that should be discussed is the units for angles and phases.

source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+      - point: [-0.088, -0.021]    # V4
+      - point: [0.023, -0.033]     # V5
+```
+This corresponds roughly to what is shown in {numref}`f:aperture`.
+
+If the boolean `absolute_vertices` is set `False`, which is the default,
+the vertex point positions are with respect to the `center` point. 
+That is, the vertex point positions in absolute terms are the positions given with each vertex plus
+the position of the `center`. If `absolute_vertices` is `True`, the vertex point positions 
+are independent of the `center` point. The default position of the `center` is the origin.
```
cemitch99, line 189: I'm not sure I understand the purpose of this flag.  When would a user need the case `absolute_vertices = True`?  Isn't this just equivalent to just using `absolute_vertices = False` together with the default `center = [0,0]`?

source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+If there is a `radius` element, there can also be a  single `tilt` element between
+the `point`s but `tilt` can only be present if there is a `radius` present.
+There can also be a `radius` element and `tilt` element after the last `point` in the `vertices`list. 
+If these are present, they are considered to be between the last and fist `point`s in the list.
+
+The aperture is constructed by connecting consecutive `point`s in the `vertices` list along with
+connecting the last `point` to the first. If there is no `radius` between consecutive `point`s,
+the aperture follows a straight line between the points. If there is a `radius` element between
+two points, the aperture is either a circular arc or a section of an ellipse. 
+A `radius` can have either a single value or a list of two values. If a `radius` has a single
```
cemitch99, line 203: Recommend specifying the units of `radius` here (m).

source/element_parameters.md
```diff
@@ -0,0 +1,225 @@
...
+value, this is the radius of the circular arc connecting the vertices. In this case, the `tilt`
+element may not be present. If the `radius` has two values, these are the half lengths of the 
+principal diameters. If the `tilt` is zero (the default), the ellipse is upright with the first
```
cemitch99, line 206: Recommend specifying the units of `tilt` here (e.g., deg or rad).

source/element_parameters.md
cemitch99: Agreed.

source/element_parameters.md
cemitch99: Yes, this one is a bit special.  Either name seems fine with me.

source/element_parameters.md
cemitch99: What about `vertex_defined`?

source/element_parameters.md
DavidSagan: Yes. If the center = [0,0] then the setting of absolute_vertices does not matter. absolute_vertices is just a convenience and not a necessity if center is not equal to [0,0].

source/element_parameters.md
DavidSagan: **All** lengths are meters as stated in the Introduction. Should that be repeated everywhere?

source/element_parameters.md
DavidSagan: Again are the units of all quantities be specified even though they are specified in the Introduction?

source/element_parameters.md
DavidSagan: Or how about `vertexes`?

cemitch99: `vertices`?

source/element_parameters.md
cemitch99: I think adding this boolean input adds an unnecessary complication, but that's just one opinion.

source/element_parameters.md
DavidSagan: OK `vertices` it is.

source/element_parameters.md
cemitch99: The fact that `radius` is a length, and therefore defined in units of m, is clear due to the naming.  However, for parameters in future groups this may not be the case.  When each list of group parameters is _first_ introduced in the document, I think it is still good to specify _either_ the units or the "category"--length, mass, angle, etc.--that will determine the units.  Something short, as in `radius [m]` or `radius [length]`.

source/element_parameters.md
DavidSagan: The advantage of a`bsolute_vertices` = True is that it is easier to visualize where the aperture is (since the aperture position is independent of `center`). The advantage of `absolute_vertices` = False is that moving the aperture as a whole is easy (just vary the `center`).  Which one would you give up if `absolute_vertices` was not present?

Note: If the aperture does not enclose the origin (0,0) then `center` (which must be inside the aperture) cannot be (0,0).

source/element_parameters.md
cemitch99: Same comment as above.

source/element_parameters.md
DavidSagan: Done.

source/element_parameters.md
DavidSagan: Done.

source/element_parameters.md
DavidSagan: Done.

source/element_parameters.md
DavidSagan: Done.

source/element_parameters.md
DavidSagan: Added units info.

source/element_parameters.md
DavidSagan: Added units info.

cemitch99 (approved):

PR #49, opened 2025-03-24
Fix MyST Markdown Syntax

EZoni:
The syntax `\vn{text}` does not seem to be rendered correctly in MyST markdown (see screenshot from https://pals-project.readthedocs.io/en/latest/introduction.html below). Was that maybe a custom LaTeX shortcut for bold font?

Screenshot from 2025-03-24 14-31-45

### DIFF
--- a/source/introduction.md
+++ b/source/introduction.md
@@ -15,7 +15,7 @@ All `keywords` in this standard are case-sensitive.
 (s:elements.intro)=
 ## Lattice Elements
 
-The basic building block used to describe an accelerator is the lattice \vn{element}. Typically,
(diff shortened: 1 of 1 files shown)


DavidSagan:
Yes `\vn{...}` was for emphasized text. So one question is do people like using bold or backticks in this case?

EZoni:
Personally I would use `bold` or `italic` if the purpose is to emphasize the word and `backticks` if the word refers to a code element (e.g., the name of a Python class, the name of a variable, etc.). But I'm fine with other choices as well, of course.

DavidSagan:
@EZoni Fine with me.

PR #51, opened 2025-03-30
Parameter devel2

DavidSagan:
More element parameter devel.


source/parameters/magneticmultipole.md
```diff
@@ -0,0 +1,89 @@
...
+MagneticMultipoleP:
+  tilt7: 0.7        - Tilt of 7th order multiple
+  Bn3: 27.3         - Normal multipole component of order 3.
+  Bn3L: 3.47e1      - Length integrated normal multipole component of order 3.
+```
+The length integrated values are related to the non-integrated values via
+```{math}
+  (B_n, B_s, K_n, K_s) = L \, (B_n, B_s, K_n, K_s)
+```
+where `L` is the length of the element.
```
EZoni, line 57: I wonder if there is a way to make the notation in this formula clearer?

Here I see $(B_n,B_s,K_n,K_s) = L(B_n,B_s,K_n,K_s)$. For one single component this reads, e.g., $B_n = L B_n$, which I think may be not too clear, given that the quantity $B_n$ appears on both sides of the equation.

Also, if the sentence "where `L` is the length of the element" refers to the equation, maybe we could write "L" in math style rather than code style.

source/element-parameters.md
```diff
@@ -66,387 +69,68 @@ Element:
 ```
 
-(s:ac.kicker.params)=
-## ACKickerP:  AC-Kicker Parameters
+## Dependent and User Set parameters.
 
-In Construction...
+Some parameters are set in the lattice file. These parameters are called "User settable".
```
cemitch99, line 74: I'm not sure about the word "settable".  I suggest calling these "User set" parameters, using the same phrase that appears in the title of this section.

source/element-parameters.md
```diff
@@ -66,387 +69,68 @@ Element:
 ```
 
-(s:ac.kicker.params)=
-## ACKickerP:  AC-Kicker Parameters
+## Dependent and User Set parameters.
 
-In Construction...
+Some parameters are set in the lattice file. These parameters are called "User settable". 
+Some parameters are not settable and will be computed by lattice expansion code.
```
cemitch99, line 75:
```suggestion
Some parameters cannot be set by the user and will be computed by lattice expansion code. 
```

source/element-parameters.md
```diff
@@ -66,387 +69,68 @@ Element:
...
-(s:ac.kicker.params)=
-## ACKickerP:  AC-Kicker Parameters
+## Dependent and User Set parameters.
 
-In Construction...
+Some parameters are set in the lattice file. These parameters are called "User settable". 
+Some parameters are not settable and will be computed by lattice expansion code. 
+For example, the `FloorP` parameters can be set for the `BeginningEle` element which is the
+first element of any branch line. For most other elements, the `FloorP` parameters can
+be calculated starting at the `BeginningEle` and working
```
cemitch99, line 78: This section ends abruptly.  Maybe it would be good to add "(... under construction)" or something similar.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
+%---------------------------------------------------------------------------------------------------
+(s:bend.params)=
+## BendP: Bend Parameters
+
+The `BendParams` stores the parameters that characterize the shape of a [`Bend`](#s:bend) element.
```
cemitch99, line 5:
```suggestion
The `BendParams` group stores the parameters that characterize the shape of a [`Bend`](#s:bend) element. 
```

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
+%---------------------------------------------------------------------------------------------------
+(s:bend.params)=
+## BendP: Bend Parameters
+
+The `BendParams` stores the parameters that characterize the shape of a [`Bend`](#s:bend) element. 
+The only relavent shape parameter that is not in the `BendParams` is the
```
cemitch99, line 6:
```suggestion
The only relevant shape parameter that is not in the `BendParams` is the
```

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+length `L` which is in the `LengthParams`.
```
cemitch99, line 7:
```suggestion
length `L` which is in the `LengthParams` group.
```

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+bend_field_ref    - Reference bend field.
+L_chord           - Chord length.
+tilt_ref          - Reference tilt.
+e1                - Entrance end pole face rotation.
+e2                - Exit end pole face rotation.
+e1_rect           - Entrance end pole face rotation.
+e2_rect           - Exit end pole face rotation.
+edge_int1         - Entrance end fringe field integral.
+edge_int2         - Exit end fringe field integral
+```
```
cemitch99, line 23: There is redundancy in this choice of parameters, which provides a lot of flexibility to the user.  However, it also requires that consistency conditions be imposed, and it complicates the description of the standard.  For example, I don't see why parameters `(e1_rect, e2_rect)` are needed.  Why not use `e1` and `e2` only, and define these with respect to the appropriate pole face associated with the given `bend_type`?  Likewise, do we need `L_chord` as a separate input if the bend `Length`, `angle`, and `1/radius` are all specified?  The same comment applies to the pair `g` and `bend_field_ref`.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+A) Bend geometry with positive bend angle. For the geometry shown,
+`g`, `angle`, `rho`, `e1`, `e2`, `e1_rect`, and `e2_rect` are all positive.
+
+B) Bend geometry with negative bend angle. For the geometry shown,
+`g`, `angle`, `rho`, `e1`, `e2`, `e1_rect`, and `e2_rect` are all negative.
+Note: The figures are drawn for zero `ref_tilt` where the rotation axis is parallel to the
+{math}`y`-axis.
+```
+
+In detail:
```
cemitch99, line 41: It would be clearer if the detailed descriptions of these parameters appear in the same order as the list appearing in lines 11-22.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+and the placement of lattice elements downstream from the bend. The actual ("total") field is
+a vector sum of
+`bend_field_ref` plus the value of the `Bn0`  and `Bs0` multipoles. If `tilt0` and `Bs0`
+are zero, the actual field is
+  ```{code} yaml
+  B-field(total) = bend_field_ref + Bn0
+  ```
+  See the discussion of `g` and `Kn0` below for more details.
+%
+- **bend_field (output param), norm_bend_field (output_param)**
```
cemitch99, line 58: Are output parameters discussed elsewhere in the standard?  This is really part of the analysis of the lattice--is it necessary or desirable to support these at the standard level?  If so, I would move them together into a separate section at the end, rather than including them in the definition of the parameter group.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+The `L_rectangle` parameter is the "rectangular" length defined to be the distance between the
+entrance and exit points. The coordinate system used for the calculation is defined by the setting
+of `fiducial_pt`. {numref}`f:rbend` shows `l_rectangle` for `fiducial_pt` set to
+`entrance_end` (the coordinate system corresponds to the entrance coordinate system of the bend).
+In this case, and in the case where `fiducial_pt` is set to `exit_end`, the rectangular
+length will be {math}`rho sinalpha`. If `fiducial_pt` is set to `none` or `center`,
+`l_rectangle` is the same as the chord length.
+%
+- **ref_tilt**
+The `ref_tilt` attribute rotates a bend about the longitudinal axis at the entrance face of the
```
cemitch99, line 168: In the list on lines 11-22, this is called `tilt_ref`.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+`g` and not `Kn0` which results in downstream elements moving around. See Sref{s:ex.chicane}
+for an example.
+
+  Note: A positive `g`, which will bend particles and the reference orbit in the {math}`-x` direction
+represents a field of opposite sign as the field due a positive `hkick`.
+%
+- **h1, h2**
+The attributes `h1` and `h2` are the curvature of the entrance and exit pole faces.
+%
+- **L, L_arc, L_chord, L_sagitta (output param)**
```
cemitch99, line 150: By `L` here, do you mean the `Length` parameter?  It was stated earlier in this section that the `Length` parameter is defined outside any parameter group.  If we want to clarify how the length is defined for a Bend element, then I think this should be placed elsewhere, outside of the parameter group description.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+%
+- **L_rectangle**
+The `L_rectangle` parameter is the "rectangular" length defined to be the distance between the
+entrance and exit points. The coordinate system used for the calculation is defined by the setting
+of `fiducial_pt`. {numref}`f:rbend` shows `l_rectangle` for `fiducial_pt` set to
+`entrance_end` (the coordinate system corresponds to the entrance coordinate system of the bend).
+In this case, and in the case where `fiducial_pt` is set to `exit_end`, the rectangular
+length will be {math}`rho sinalpha`. If `fiducial_pt` is set to `none` or `center`,
+`l_rectangle` is the same as the chord length.
+%
```
cemitch99, line 166: There is a lot of redundancy in this list of _length_ parameters.  It would be good to clarify which of these parameters are inputs in the `BendP` parameter group, and which are intended as "output parameters".

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+- **g, rho (output param)**
+The Reference bending radius which determines the reference coordinate system is `rho` (see
+[](#s:ref)). `g` = `1/rho` is the "bend strength" and is proportional to the Reference
+dipole magnetic field. `g` is related to the reference magnetic field `bend_field_ref` via
+  ```{math}
+  :label: gqpb
+
+  \text{g} = \frac{q}{p_0} \cdot \text{bend_field_ref}
+end{equation}
```
cemitch99, line 128:
```suggestion
\end{equation}
```

cemitch99: This `end{equation}` appears as text in the rendered document.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+`ref_tilt`. Bends in Bmad do not have a `tilt` attribute.
+
+  Important! Do not use `ref_tilt` when doing misalignment studies for a machine. Trying to misalign
+a dipole by setting `ref_tilt` will affect the positions of all downstream elements! Rather, use the
+`tilt` parameter.
+
+%---------------
+
+  The attributes `g`, `angle`, and `L` are mutually dependent. If any two are specified for
+an element AcceleratorLattice will calculate the appropriate value for the third.
```
cemitch99, line 182: What is the object `AcceleratorLattice` that is referenced here?  Is this defined elsewhere, or does this just refer to some logic in the Translator layer?

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+
+  In the local coordinate system ([](#s:ref)), looking from "above" (bend viewed from positive
+{math}`y`), and with `ref_tilt` = 0, a positive `angle` represents a particle rotating clockwise. In
+this case. `g` will also be positive. For counterclockwise rotation, both `angle` and `g`
+will be negative but the length `l` is always positive. Also, looking from above, a positive
+`e1` represents a clockwise rotation of the entrance face and a positive `e2` represents a
+counterclockwise rotation of the exit face. This is true irregardless of the sign of `angle` and
```
cemitch99, line 189:
```suggestion
counterclockwise rotation of the exit face. This is true independently of the sign of `angle` and
```

source/parameters/magneticmultipole.md
cemitch99: Maybe the length-integrated values could be explicitly defined by `BnNL` = _L_*`BnN`, or something similar.

source/parameters/referencechange.md
```diff
@@ -0,0 +1,48 @@
+%---------------------------------------------------------------------------------------------------
+(s:ref.change.params)=
+## ReferenceChangeP: Change in Reference Parameters
```
cemitch99, line 3: Parameter groups that explicitly refer to "reference" orbit values will stimulate a lot of discussion.  There has been a stated goal of separating the lattice description from the dynamics description, and here these are explicitly mixed.

source/figures/bend.svg
cemitch99, line 1: Is it really necessary to have both Figures A) and B) here?  I understand the motivation, but this may cause more confusion than clarification.  I would suggest keeping Figure A) only, possibly without the labels defining `e1_rect` and `e2_rect`.

cemitch99: Actually, maybe it would make sense to replace Figure B) with an upright figure illustrating the corresponding definitions for a bend of type RECTANGULAR.

source/element-parameters.md
DavidSagan: The problem with "User set" is that this implies that the User has set the parameter. As opposed to "User settable" which implies that the User **may** set the parameter.

source/element-parameters.md
DavidSagan: changed to:
```
Some parameters will be computed by the Translator during lattice expansion. Such parameters
may not be set by the user.
```

DavidSagan: This brings up the question whether it is an error for Translator computed quantities to be present in the lattice file. I can see a use case where the User wants to print out the Translated file complete with all computed quantities.

source/parameters/bend.md
DavidSagan: As you note, a user may have a rectangular bend and may want to therefore specify `e1_rect`, `e2_rect`, and `L_chord`. That said, it would indeed simplify things if these quantities where made to be not User settable. This should be discussed.

source/parameters/bend.md
DavidSagan: This needs to be discussed. There can be a third category which are parameters that can be input or output.

source/parameters/magneticmultipole.md
DavidSagan: Yes this was a typo with the `L` left off. This has been fixed

source/element-parameters.md
DavidSagan: Added "and working forward computing the floor parameters
element-by-element."

source/parameters/bend.md
DavidSagan: Done.

source/parameters/bend.md
DavidSagan: Fixed.

source/parameters/bend.md
DavidSagan: The parameter list in lines 11-22 has been reordered.

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+all downstream lattice elements
+unchanged but will vary a particle's orbit. One common mistake when designing lattices is to vary
+`g` and not `Kn0` which results in downstream elements moving around. See Sref{s:ex.chicane}
+for an example.
+
+  Note: A positive `g`, which will bend particles and the reference orbit in the {math}`-x` direction
+represents a field of opposite sign as the field due a positive `hkick`.
+%
+- **h1, h2**
+The attributes `h1` and `h2` are the curvature of the entrance and exit pole faces.
```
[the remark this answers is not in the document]
DavidSagan: Fixed.

source/parameters/bend.md
DavidSagan: This should be `Length`. Fixed.

source/parameters/bend.md
DavidSagan: Changed everything to `tilt_ref`.

source/parameters/bend.md
DavidSagan: Fixed.

source/parameters/bend.md
DavidSagan: Changed.

source/parameters/bend.md
DavidSagan: Typo. Fixed.

source/overview.md
```diff
@@ -19,12 +19,59 @@ This custom information is generally not usable by other programs but can be use
 lattice files that it generated. 
 
 
+%---------------------------------------------------------------------------------------------------
 ## What PALS Is
 
-PALS is a schema that defines things like the names of various lattice element types, how to organize lattice
-elements into lines which beams of particles or X-rays can move through, etc. 
+PALS is a schema that defines things like the names of various lattice element types, 
+how to organize lattice elements into lines which beams of particles or X-rays can move through, etc.
```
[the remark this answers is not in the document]
ChristopherMayes: `X-rays` is too specific. Could we call it `photons` instead?

source/overview.md
```diff
@@ -19,12 +19,59 @@ This custom information is generally not usable by other programs but can be use
...
+:width: 70%
+:name: f:trans
+
+Schematic showing how a simulation program can interact with a PALS complient lattice file.
+```
+The components here are:
+- **Simulation program:** A simulation program.
+- **Interface layer:** Code to interface between the structures defined by the program and the 
+lattice structure defined by the Translator. The program maintainer(s) must create the interface 
+layer since this code will be program specific. However, structure translation should (hopefully) be
```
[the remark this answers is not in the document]
DavidSagan: Changed to `This code will be program specific and outside the
scope of the standard.`

source/parameters/bend.md
```diff
@@ -0,0 +1,193 @@
...
+bend_type         - Type of bend. Default: SECTOR.
+angle             - Reference bend angle.
+g                 - Reference bend strength = 1/radius.
+bend_field_ref    - Reference bend field.
+L_chord           - Chord length.
+tilt_ref          - Reference tilt.
+e1                - Entrance end pole face rotation.
+e2                - Exit end pole face rotation.
+e1_rect           - Entrance end pole face rotation.
+e2_rect           - Exit end pole face rotation.
```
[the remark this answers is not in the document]
DavidSagan: Added `with respect to a sector geometry.` and `with respect to a rectangular geometry.`

source/parameters/aperture.md
```diff
@@ -0,0 +1,192 @@
...
+the position of the `center`. If `absolute_vertices` is `True`, the vertex point positions 
+are independent of the `center` point. The default position of the `center` is the origin.
+
+The `list` component of `vertices` contains an ordered  list of vertex 
+`point`s with each one specifying an {math}`(x, y)` position. 
+In between any two `point`s, there can optionally be a single `radius` element.
+If there is a `radius` element, there can also be a  single `tilt` element between
+the `point`s but `tilt` can only be present if there is a `radius` present.
+There can also be a `radius` element and `tilt` element after the last `point` in the `list`. 
+If these are present, they are considered to be between the last and fist `point`s in the list.
```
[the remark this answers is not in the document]
DavidSagan: Changed.

source/parameters/aperture.md
```diff
@@ -0,0 +1,192 @@
...
+### material
+
+The `material` parameter sets the material of the aperture. 
+Using chemical formulas like `Cu` and `Fe` are the most portable.
+
+### vertices component
+
+The `VERTICES` setting for `shape` is for defining an aperture using a 
+set of vertex points as illustrated in {numref}`f:aperture`B. 
+Between vertex points, the aperture can can follow a straight line or the arc of an ellipse.
```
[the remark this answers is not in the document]
DavidSagan: Changed.

source/parameters/aperture.md
```diff
@@ -0,0 +1,192 @@
+%---------------------------------------------------------------------------------------------------
+(s:aperture.params)=
+## ApertureP: Aperture Parameters
+
+The `ApertureP` parameter group contains parameters for describing an aperture. 
+The components of this group are:
+```{code} yaml
+x_limit                      # [m] Vector of two real numbers
+y_limit                      # [m] Vector of two real numbers
```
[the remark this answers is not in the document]
DavidSagan: Changed.

source/overview.md
```diff
@@ -19,12 +19,59 @@ This custom information is generally not usable by other programs but can be use
...
+Schematic showing how a simulation program can interact with a PALS complient lattice file.
+```
+The components here are:
+- **Simulation program:** A simulation program.
+- **Interface layer:** Code to interface between the structures defined by the program and the 
+lattice structure defined by the Translator. The program maintainer(s) must create the interface 
+layer since this code will be program specific. However, structure translation should (hopefully) be 
+relatively straight forward.
+- **Translator:** Package supplied with the lattice standard. The Translator has code to put 
+lattice information into a "expanded form".
```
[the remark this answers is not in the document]
DavidSagan: Changed.

source/overview.md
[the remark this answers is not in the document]
DavidSagan: Changed to photons. Can you suggest an alternative phrasing for `which beams of particles or X-rays can move through.`?

source/figures/bend.svg
DavidSagan: What exactly a negative angle means can be a source of confusion so figure B is useful.

source/parameters/bend.md
DavidSagan: Yes the discussion was expanded in the section `User Settable (Input) and Dependent (Output) parameters.` And it is important to have these. More work is need on documenting which parameters are input and which are output but that is for a later PR.

source/parameters/referencechange.md
DavidSagan: Yes but for `normalized` parameters like `Kn1`, the reference energy is needed to convert to a field.

source/parameters/aperture.md
```diff
@@ -0,0 +1,192 @@
...
+  5) Check downstream aperture if there is one.
+  6) Convert from body coordinates to branch coordinates.
+  7) End at downstream end of element.
+```
+
+### material
+
+The `material` parameter sets the material of the aperture. 
+Using chemical formulas like `Cu` and `Fe` are the most portable.
+
```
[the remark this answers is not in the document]
DavidSagan: I added the following: `The difference between the two lists is simply that lines 2 and 3 are switched along with lines 5 and 6.`

source/parameters/aperture.md
```diff
@@ -0,0 +1,192 @@
...
+```{code} YAML
+  1) Start at upstream end of element
+  2) Convert from branch coordinates to body coordinates.
+  3) Check upstream aperture if there is one.
+  4) Track through the element body.
+  5) Check downstream aperture if there is one.
+  6) Convert from body coordinates to branch coordinates.
+  7) End at downstream end of element.
+```
+
```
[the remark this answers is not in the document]
DavidSagan: The difference between the two lists is simply that lines 2 and 3 are switched along with lines
5 and 6.

source/parameters/bend.md
cemitch99: I think this comment and the discussion that followed (regarding parameters specifying a Bend) may not yet be fully resolved.  This stimulated discussion about RBENDs and an exchange with Laurent Deniau, but I'm not sure how this was left.

source/parameters/bend.md
cemitch99: I don't think this has been discussed yet.

source/parameters/aperture.md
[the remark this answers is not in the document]
cemitch99: "The use of chemical formulas like `Cu` and `Fe` is the most portable option."

source/parameters/bend.md
DavidSagan: The discussion about what is User settable and what is not is mostly being left to later PRs.

source/parameters/bend.md
DavidSagan: The discussion about what is User settable and what is not is mostly being left to later PRs.

source/parameters/aperture.md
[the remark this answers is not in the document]
DavidSagan: @cemitch99 But `formulas` is plural.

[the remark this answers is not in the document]
cemitch99: @DavidSagan But `formulas` is not the subject of this sentence.  The subject is singular: `use` (as a noun), and so is the verb `is`.  The prepositional phrase `of chemical formulas like...` is a modifier of `use`.

source/parameters/bend.md
cemitch99: Thanks.  Just to clarify -- this comment is about the parameters needed to specify a Bend (and the case of RBENDS in particular) -- separate from the discussion about what is User settable.

DavidSagan: Note that as a result of the conversation with Laurent, I did change the PR to use `SBend` and `RBend` elements instead of just having `Bend` elements.

source/parameters/aperture.md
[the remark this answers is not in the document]
DavidSagan: OK changed.

PR #62, opened 2025-06-22
Sphinx LaTeX: Support SVG

ax3l:
The readthedocs builder is set right now to automatically build PDFs via LaTeX. LaTeX does not support SVGs directly, but other vector formats like PDF. There is a sphinx plugin that can [auto-convert SVGs to PDFs](https://github.com/missinglinkelectronics/sphinxcontrib-svg2pdfconverter) (or [rasterized images](https://www.sphinx-doc.org/en/master/usage/extensions/imgconverter.html#module-sphinx.ext.imgconverter)) via, e.g., Inkscape. Trying to use that now.

For source code / repo purposes, SVGs are superior to PDFs, so we do not want to change our source images (in git).

### DIFF
--- a/conda.yaml
+++ b/conda.yaml
@@ -22,5 +22,6 @@ dependencies:
   - sphinx-copybutton
   - sphinx-design
   - sphinxcontrib-bibtex
+  - sphinxcontrib-svg2pdfconverter-with-all
   - sphinxcontrib-napoleon
 # - sphinx_rtd_theme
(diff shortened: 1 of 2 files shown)


ax3l:
I can only see if this works after merge, unfortunately. (Because we only build an HTML preview in the PR.)

DavidSagan:
@ax3l Fine with me for you to merge, test and then revert if necessary. No need for further review.

PR #65, opened 2025-06-30
Fix Read the Docs builds

EZoni:
Our Read the Docs builds are failing:

Screenshot from 2025-06-30 10-54-39

This explains why our online documentation is not up-to-date.

Interestingly, this occurrs only for commits in `main`, but not in individual PRs.

The logs point to something like
```
! Missing } inserted.
<inserted text> 
                }
l.1537 {2 , B_{y0}^2}\end{split}
                                
(That makes 100 errors; please try again.)
!  ==> Fatal error occurred, no output PDF file produced!
```

This PR should fix the syntax error. 

I also remove the obsolete workflow file related to Curvenote, which we don't use anymore.


source/parameters/bend.md
```diff
@@ -85,8 +85,7 @@ Zero `e1_rect` and `e2_rect` gives a rectangular magnet shape.
 - **edge_int1, edge_int2**
 The field integral for the entrance pole face `edge_int1` is given by
   ```{math}
-  \text{edge_int1} = \int_{pole} \!\! ds \frac{B_y(s) (B_{y0} - B_y(s))}
-  {2 , B_{y0}^2}
+  \text{edge_int1} = \int_{pole} \!\! ds \frac{B_y(s) (B_{y0} - B_y(s))}{2 B_{y0}^2}
```
EZoni, line 88: @DavidSagan 

I'm bringing the components of the fraction onto the same line, in case this breaks the LaTeX build. 

Was the `,` between `2` and `B_{y0}^2` intentional? I wonder if this may contribute to breaking the build.

DavidSagan: The comma should not have been there. It accidentally got left in in the translation from latex.

EZoni:
The build once the commit is merged in `main` is still failing. I will look again at the logs and see if there's something else to fix. No idea why the build passes in the PR...

EZoni:
> It looks like the problem is with the PDF generation. See PR https://github.com/campa-consortium/pals/pull/63. Perhaps just generating HTML for now would be the best?

Oh okay, so you were already aware of this. I'll review that PR now - we can move the conversation there.

PR #72, opened 2025-07-21
Element Parameters: Update to new Structure

ax3l:
Update the Element Parameters page to the new structure of name first, then element as dict inside.

Also clarifies that we can inherit by element parameter group name and that in-element parameter groups have the name of the element they are in.

[See the preview here.](https://pals-project--72.org.readthedocs.build/en/72/element-parameters.html)

### DIFF
--- a/source/element-parameters.md
+++ b/source/element-parameters.md
@@ -2,7 +2,7 @@
 # Element Parameters
 
 Lattice elements parameters are organized into **parameter groups**. 
-All groups are organized as dictionaries (structures).
+All groups are organized as dictionaries (structures) and lists.
 At the top level, there are the groups with names like 
 `MagneticMultipoleP`, `ElectricMultipoleP`, `MetaP`, `AlignmentP`, etc. 
 By convention, group names use upper camel case and it is highly recommended that this convention


source/element-parameters.md
```diff
@@ -32,39 +32,39 @@ Any group can be given a **name** and the values can be used in another group of
...
   ApertureP:
     inherit: q1.ApertureP
 ```
 
 For an element to inherit all parameter groups from another element, just inherit the element itself:
 ```{code} yaml
-Quatrupole:
-  name: q3
+q3:
+  kind: Quatrupole  # TODO: do we need to list the kind in inheritance again?
```
ax3l, line 67: Memo to myself: Same TODO as on `ap2`. Update the code here once we clarified the above one.

ax3l, line 67:
```suggestion
  kind: Quadrupole  # TODO: do we need to list the kind in inheritance again?
```

source/element-parameters.md
```diff
@@ -32,39 +32,39 @@ Any group can be given a **name** and the values can be used in another group of
 using **import**.
 For example:
 ```{code} yaml
-ApertureP:
-  name: ap1
+ap1:
+  kind: ApertureP
```
DavidSagan, line 36: Hmmm. I'm rethinking giving names to parameter groups. As an alternative, parameter groups could be identified using a `<element-name>.<parameter-group-name>` syntax. For example, if an  `ApertureP` group is used in an element named `Q1` then the group could be referred to as `Q1.ApertureP`. We should discuss this at the next meeting.

ax3l: To implement what you are suggesting, we could remove the first two ap1/ap2 examples and just stay with the q1/q2 example?

ax3l: I pushed a change now that adds examples for both.

source/element-parameters.md
```diff
@@ -32,39 +32,39 @@ Any group can be given a **name** and the values can be used in another group of
...
   ApertureP: 
     x_limit: [-0.03, 0.04]
     y_limit: [-0.02, 0.03]
 ```
 And an element can inherit a parameter group from another element:
 ```{code} yaml
-Quatrupole:
-  name: q2
+q2:
+  kind: Quadrupole
```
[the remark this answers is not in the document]
ax3l: Ok, had the same feeling. Then we can simplify these examples and omit it.

source/element-parameters.md
```diff
@@ -19,10 +19,10 @@ and `s_position` (the longitudinal position of the element).
...
 the following is not allowed:
 ```{code} yaml
-Quadrupole:
-  name: q10w
-  MetaP: ...
-  MetaP: ...    # Second instance not allowed!
+q10w:
+  kind: Quadrupole
+  ApertureP: ...
+  ApertureP: ...    # Second instance not allowed!
```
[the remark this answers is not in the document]
ax3l: Synthactically, as a YAML element (dict): not really.

But in the context of PALS, where we always put an element (dict) into a list, definitely.

We could make this simple (focus on the element) as I have drafted now, or make it easy for people to copy-paste into new lattice files (with the `-` already added), as you suggested. I am ok for the latter.

source/element-parameters.md
```diff
@@ -2,7 +2,7 @@
 # Element Parameters
 
 Lattice elements parameters are organized into **parameter groups**. 
-All groups are organized as abstract syntax trees.
+All groups are organized as dictionaries (structures).
```
ax3l, line 5:
```suggestion
All groups are organized as dictionaries (structures) and lists.
```

source/element-parameters.md
DavidSagan: Also that should be `Quadrupole`.

source/element-parameters.md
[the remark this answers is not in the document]
DavidSagan: Note: Since the standard allows modifying element parameters after an element is defined, order is important.

PR #76, opened 2025-08-25
Added misalignment documentation.

DavidSagan:
Added misalignment documentation along with numerous small tweaks.


source/element-parameters.md
```diff
@@ -26,6 +21,26 @@ the following is not allowed:
...
+
+There are five element parameters that are so common, and do not fit into
+any of the parameter groups, that they are not grouped. 
+These element parameters are:
+```{code} yaml
+  field_master: NotSet  # [Boolean] See Below.
+  is_on: T              # [Bollean] Turns on or off the fields in an element. When off, the element looks like a drift.
+  length: 0             # [m] Length of element. For bends this is the arc length.
+  name: ""              # [string] The name of element.
+  s_position: 0         # [m] The longitudinal position of the element.
```
cemitch99, line 35: Does "kind" belong here?

source/parameters/aperture.md
```diff
@@ -4,16 +4,16 @@
...
   shape: ""                       # [string] Aperture shape switch
-  location: ENTRANCE_END          # [string] Aperture location switch
-  aperture_shifts_with_body: ...  # [Boolean] ... TODO: describe ... TODO: default
-  vertices: {}                    # [Dictionary] ... TODO: describe ...
+  location: ENTRANCE_END          # [enum] Aperture location switch
+  vertices: []                    # [array] Array of vertex points. See below.
   material: ""                    # [string] Material of the Aperture
   thickness: 0                    # [m] Real number
+  aperture_shifts_with_body: F    # [Boolean] See below.
+  aperture_active: T              # [Boolean] F implies aperture is not operating.
```
ax3l, line 16:
```suggestion
  aperture_shifts_with_body: false  # [Boolean] See below.
  aperture_active: true             # [Boolean] F implies aperture is not operating.
```

source/element-parameters.md
```diff
@@ -26,6 +21,27 @@ the following is not allowed:
...
 %---------------------------------------------------------------------------------------------------
+(s:non.params)=
+## Non-Parameter Group Parameters
+
+There are element parameters that are so common, and do not fit into
+any of the parameter groups, that they are not grouped. 
+These element parameters are:
+```{code} yaml
+  field_master: NotSet  # [Boolean] See Below.
+  is_on: T              # [Bollean] Turns on or off the fields in an element. When off, the element looks like a drift.
```
ax3l, line 32:
```suggestion
  is_on: true           # [Boolean] Turns on or off the fields in an element. When off, the element looks like a drift.
```

source/lattice-construction.md
```diff
@@ -17,7 +17,7 @@ Example:
             periodic: True
```
ax3l, line 17:
```suggestion
            periodic: true
```

ax3l (changes requested):
Thank you for the updates :+1: 

Small wish to not reinvent the wheel on Boolean values and keep it simple, so we do not need to write custom parsers for them in implementations (since we use YAML syntax for documentation of the standard, they are well-defined already).

source/lattice-construction.md
DavidSagan: Corrected.

PR #79, opened 2025-09-30
Matter Element and MaterialP

danielkallendorf:
### Matter
Provide a proposed name for the lattice element.  This can be an abbreviation, and should be at most 1-2 words (e.g., Quadrupole).

## Matter Element (replaces foil)

Materials which fully occupy the beamline, like targets, (stripper-)foils, vacuum windows, gas cells or degraders.
This element can cause energy-loss, angualar and energy straggling, as well as change of charge state or particle type.

## MaterialP:  Definition of materials

The definition of materials is inspired by Geant4's material handling. One can either use the name, which is the symbol of an element (`"G4_Cu"`, `"G4_H"`, ...) or compound (`"G4_STAINLESS-STEEL"`).
In the future all elements and materials in the [G4 Manual](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/Appendix/materialNames.html) will be supported.

Alternatively custom elements can be defined by

```yaml
ElementP:
- name:      # [string] The name of the material
- density:   # [g/cm^3] The density of the material
- Z:         # [int]    The atomic number
- A:         # [int]    The mass number (equals Z+N)
- N:         # [int]    The neutron number
- m:         # [g/mol]  The atomic mass
```

and compound by

```yaml
CompoundP:
- name:      # [string] The name of the material
- density:   # [g/cm^3] The density of the material
- elements:  # List of MaterialP
- ratio:     # List of ratios of the elements
```

### Physical model
I assume a material should be  described precise enough by state, density, thickness and composition.

Note: Due to the stochastic nature of beam-matter interaction, the Taylor-/transfer-map-formalism can only account for the mean average effect.

### DIFF
--- a/source/lattice-element-kinds.md
+++ b/source/lattice-element-kinds.md
@@ -330,27 +330,6 @@ Element parameter groups associated with this element kind are:
 
 
 %---------------------------------------------------------------------------------------------------
-(s:foil)=
-### Foil Element
-
-A `Foil` element represents a planar sheet of material which can strip electrons from a particle. In
-conjunction, there will be scattering of the particle trajectory as well as an associated energy loss.
-Material that can strip electrons from a particle
-will also cause energy loss and diffusion.
-
-Under Construction...
-
-Element parameter groups associated with this element kind are:
-- [**ApertureP**](#s:aperture.params): Aperture parameters.
-- [**BodyShiftP**](#s:bodyshift.params): Orientation of element with respect to its nominal position.
-- [**FloorP**](#s:floor.params): Floor position and orientation.
-- [**FoilP**](#s:foil.params): Foil parameters.
-- [**MetaP**](#s:meta.params): Meta parameters.
-- [**ReferenceP**](#s:ref.params): **Output Parameters.** Reference parameters.
-- [**TrackingP**](#s:tracking.params): Tracking parameters.
-
-
-%---------------------------------------------------------------------------------------------------
 (s:fork)=
 ### Fork Element
 
@@ -486,6 +465,24 @@ Element parameter groups associated with this element kind are:
 - [**TrackingP**](#s:tracking.params): Tracking parameters.
 
 The length of this element is considered to be zero so if `length` is specified, it must be zero.
+
+
+%---------------------------------------------------------------------------------------------------
+(s:matter)=
+## Matter Element
+
+Materials which fully occupy the beamline, like targets, (stripper-)foils, vacuum windows, gas cells or degraders.
+This element can cause energy-loss, angualar and energy straggling, as well as change of charge state or particle type.
+
+Element parameter groups associated with this element kind are:
+- [**ApertureP**](#s:aperture.params): Aperture parameters of the outer (blocking) aperture.
+- [**BodyShiftP**](#s:bodyshift.params): Orientation of element with respect to its nominal position.
+- [**FloorP**](#s:floor.params): Floor position and orientation.
+- [**MetaP**](#s:meta.params): Meta parameters.
+- [**ReferenceP**](#s:ref.params): Reference parameters.
+- [**ReferenceChangeP**](#s:refchange.params): Reference energy change and/or reference time correction.
+- [**TrackingP**](#s:tracking.params): Tracking parameters.
+- [**MatterP**](#s:matter.params): Matter parameters.
 
 
 %---------------------------------------------------------------------------------------------------
--- a/source/lattice-element-kinds.md
+++ b/source/lattice-element-kinds.md
@@ -484,7 +484,6 @@ Element parameter groups associated with this element kind are:
 - [**TrackingP**](#s:tracking.params): Tracking parameters.
 - [**MatterP**](#s:matter.params): Matter parameters.
 
-
 %---------------------------------------------------------------------------------------------------
 (s:multipole)=
 ### Multipole Element
--- a/source/parameters/aperture.md
+++ b/source/parameters/aperture.md
@@ -16,7 +16,7 @@ ApertureP:
   shape: ELLIPTICAL                # [enum] Aperture shape switch
   location: ENTRANCE_END           # [enum] Aperture location switch
   vertices: []                     # [array] Array of vertex points. See below.
-  material: ""                     # [string] Material of the Aperture
+  material: ""                     # [MaterialP] Material of the Aperture
   thickness: 0                     # [m] Real number
   aperture_shifts_with_body: false # [logical] See below.
   aperture_active: true            # [logical] false implies aperture is not operating.
@@ -125,7 +125,6 @@ The difference between the two lists is simply that lines 2 and 3 are switched a
 ### material
 
 The `material` parameter sets the material of the aperture. 
-Using chemical formulas like `Cu` and `Fe` is the most portable.
 
 ### vertices component
 
--- a/source/parameters/material.md
+++ b/source/parameters/material.md
@@ -14,7 +14,7 @@ ElementP:
 - A:         # [int]    The mass number (equals Z+N)
 - N:         # [int]    The neutron number
 - m:         # [g/mol]  The atomic mass
-```{code} yaml
+```
 
 and compunds by
 
@@ -24,7 +24,7 @@ CompundP:
 - density:   # [g/cm^3] The density of the material
 - elements:  # List of MaterialP
 - ratio:    # List of ratios of the elements
-```{code} yaml
+```
 
 ### Examples
 `MaterialP: "G4_Cu"` or `MaterialP: "G4_STAINLESS-STEEL"` or define a new material.
@@ -37,7 +37,7 @@ MaterialP:
 - Z: 1
 - A: 1
 - m: 2.01
-```{code} yaml
+```
 and deuterated polyethylene (C2H4) can be defined based on the previous definition
 ```{code} yaml
 MaterialP:
@@ -49,6 +49,6 @@ MaterialP:
 - ratio:
   - 2
   - 4
-```{code} yaml
+```
 
 
--- a/source/parameters/material.md
+++ b/source/parameters/material.md
@@ -38,7 +38,7 @@ If density is not given it will be inferred form the densities and mass fraction
 
 
 ### Examples
-`MaterialP: "Cu"` or `MaterialP: "G4_STAINLESS-STEEL"` or define a new material.
+`material: "Cu"` or `material: "G4_STAINLESS-STEEL"` or define a new material.
 
 Liquid deuterium can be defined with the `ElementP` parameters:
 ```{code} yaml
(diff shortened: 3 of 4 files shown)


DavidSagan:
Another issue is that having a single `Matter` element to represent foils, gas cells, etc. to my mind lumps too many different types of objects under one name. I would rather see separate elements. Especially since the parameter set needed for these different types will undoubtedly be different.

DavidSagan:
-  You don't need atomic mass directly. EG: `#3He` is good enough.
- OpenPMD does not define materials yet. I propose adding to OpenPMD a materials list and then PALS can inherit from that. Much better to have everything is in one place.
- in terms of element parameters, there could be something like gas pressure that is needed. We should make a survey to see what people need.

Also: Vacuum windows was mentioned. This should definitely be in a category by itself.

source/parameters/material.md
```diff
@@ -0,0 +1,65 @@
...
+### Examples
+`material: "Cu"` or `material: "G4_STAINLESS-STEEL"` or define a new material.
+
+Liquid deuterium can be defined with the `ElementP` parameters:
+```{code} yaml
+ElementP:
+- name: "Deuterium(l)"
+- density: 160
+- Z: 1
+- A: 1
```
DavidSagan: Deuterium has A = 2.

source/parameters/aperture.md
```diff
@@ -10,7 +10,7 @@ ApertureP:
   shape: ""                        # [string] Aperture shape switch
   location: ENTRANCE_END           # [enum] Aperture location switch
   vertices: []                     # [array] Array of vertex points. See below.
-  material: ""                     # [string] Material of the Aperture
+  material: ""                     # [MaterialP] Material of the Aperture
```
ax3l: Since it is not a string anymore:
```suggestion
  material: null                   # [MaterialP] Material of the Aperture
```

source/parameters/matter.md
```diff
@@ -0,0 +1,10 @@
+(s:matter.params)=
+## MatterP:  Definition of materials
```
ax3l: I am confused, `MatterP` is not referenced anywhere.

How does it relate to `MaterialP`/`ElementP`?

ax3l: Ah, there is a typo below...

source/element-kinds.md
```diff
@@ -358,6 +338,23 @@ Element parameter groups associated with this element kind are:
...
+## Matter Element
+
+Materials which fully occupy the beamline, like targets, (stripper-)foils, vacuum windows, gas cells or degraders.
+This element can cause energy-loss, angualar and energy straggling, as well as change of charge state or particle type.
+
+Element parameter groups associated with this element kind are:
+- [**ApertureP**](#s:aperture.params): Aperture parameters of the outer (blocking) aperture.
+- [**BodyShiftP**](#s:bodyshift.params): Orientation of element with respect to its nominal position.
+- [**FloorP**](#s:floor.params): Floor position and orientation.
+- [**MetaP**](#s:meta.params): Meta parameters.
```
ax3l: I assume adding the `MatterP` paramter group is the whole idea of this element:
```suggestion
- [**MatterP**](#s:matter.params): Matter parameters.
- [**MetaP**](#s:meta.params): Meta parameters.
```

source/element-kinds.md
```diff
@@ -358,6 +338,23 @@ Element parameter groups associated with this element kind are:
...
+- [**ReferenceP**](#s:ref.params): Reference parameters.
+- [**ReferenceChangeP**](#s:refchange.params): Reference energy change and/or reference time correction.
+- [**TrackingP**](#s:tracking.params): Tracking parameters.
+- [**MaterialP**](#s:material.params): Matter parameters.
```
ax3l: Ah here it is. Then just alphabetic ordering :)
```suggestion
```

I think you mean `MatterP`, not `MaterialP`.

source/parameters/matter.md
ax3l: I would propose to add the  `state` and `matter:` (formerly: `material:`) properties here and remove the `Material` concept.

ax3l:
Ah, good point.

DavidSagan:
@danielkallendorf After thinking it over, to answer your question, I would advocate that things like atomic mass be documented `output` parameters (#107) and one of the aims of the PALS C++ interface I am involved with will be to output this info with the expanded lattice.

EZoni:
To: @danielkallendorf 
Cc: @DavidSagan @ax3l 

What is the status of this PR? I see a few changes requested by @ax3l that may need to be addressed.

DavidSagan:
@danielkallendorf The history for this PR looks like a mess. I asked Claude to explain and here is the answer:

⏺ Short answer

  The PR's actual content change is 6 files / 341 lines. The 70-file, 3812-add / 50067-delete diff GitHub shows is an artifact of the
  branch's history being rewritten.

  What happened
  
  - PR #79 is from danielkallendorf/pals, branch main (their fork's default branch), opened 2025-09-30.
  - The merge base with pals-project/pals is stuck at 9cedd78 — "RTD: Skip PDF (#81)", 2025-10-15.
  - Today the author brought their branch up to date by rebasing rather than merging. You can see it in the commit list: ~107 upstream
  commits (#83, #88, #91 … #221) all appear in the PR with committer date 2026-07-23 and fresh SHAs. They're byte-identical in content
  to upstream's commits but are different commits as far as git is concerned.
  - So GitHub can't advance the merge base past Oct 2025. ahead_by: 127, behind_by: 101 — it thinks the branch simultaneously added 127
  commits and is missing 101, when really it's ~20 real commits plus 107 duplicates of work already in main.
  - The PR diff is therefore head vs. October 2025 main, i.e. nine months of upstream doc restructuring re-attributed to this PR:
  introduction.md → fundamentals.md, element-kinds.md → lattice-element-kinds.md, libraries.md → code-packages.md, the new notation.md
  / extensions.md / dispersion.md, and so on.
  - The 50k deletions are almost entirely SVG re-exports done upstream — translator.svg (−10232), fork.svg (−8955), patch-between.svg
  (−7707), ele-coord-frame.svg (−7356). None of that is this author's work.

  The real diff

  source/element-parameters.md         | 153 +++++  (stale — deleted upstream in #191)
  source/lattice-element-kinds.md      |  55 ++---
  source/lattice-element-parameters.md |  67 +++++
  source/parameters/aperture.md        |   3 +-
  source/parameters/material.md        |  65 +++++
  source/parameters/matter.md          |  10 +++
  6 files changed, 341 insertions(+), 12 deletions(-)

  One thing to flag in review: source/element-parameters.md was removed upstream when the element/parameter docs were reorganized
  (#191), and this branch re-adds it with 153 lines. That's leftover from the rebase, not intended content — the same material
  presumably belongs in lattice-element-parameters.md.

  How they'd fix the view

  Rebasing again won't help; it re-duplicates. Cleanest is to rebuild the branch on top of current main and carry over only the real
  delta:

  git fetch upstream
  git checkout -b matter-materialp upstream/main
  git checkout <their-head-sha> -- source/lattice-element-kinds.md \
      source/lattice-element-parameters.md source/parameters/aperture.md \
      source/parameters/material.md source/parameters/matter.md
  # then fold element-parameters.md's content into lattice-element-parameters.md

  and repoint the PR at that branch (also worth doing since PRs from a fork's main cause exactly this kind of trouble). The PR
  currently shows mergeable: false / dirty, so it needs conflict resolution regardless.

danielkallendorf:
@DavidSagan Yes. I messed up the rebase on the first try. I will try again.

DavidSagan:
@danielkallendorf Are you going to remove `ElementP` and use the OpenPMD standard?

source/lattice-element-kinds.md
```diff
@@ -488,6 +467,23 @@ Element parameter groups associated with this element kind are:
...
+Materials which fully occupy the beamline, like targets, (stripper-)foils, vacuum windows, gas cells or degraders.
+This element can cause energy-loss, angualar and energy straggling, as well as change of charge state or particle type.
+
+Element parameter groups associated with this element kind are:
+- [**ApertureP**](#s:aperture.params): Aperture parameters of the outer (blocking) aperture.
+- [**BodyShiftP**](#s:bodyshift.params): Orientation of element with respect to its nominal position.
+- [**FloorP**](#s:floor.params): Floor position and orientation.
+- [**MetaP**](#s:meta.params): Meta parameters.
+- [**ReferenceP**](#s:ref.params): Reference parameters.
+- [**ReferenceChangeP**](#s:refchange.params): Reference energy change and/or reference time correction.
```
DavidSagan, line 483: Also reference species change.

source/parameters/aperture.md
```diff
@@ -16,7 +16,7 @@ ApertureP:
   shape: ELLIPTICAL                # [enum] Aperture shape switch
   location: ENTRANCE_END           # [enum] Aperture location switch
   vertices: []                     # [array] Array of vertex points. See below.
-  material: ""                     # [string] Material of the Aperture
+  material: ""                     # [MaterialP] Material of the Aperture
```
DavidSagan, line 19: Better would be:
```
MaterialP: null                                  # [MaterialP] Material of the Aperture
```
a problem with using `material` is that it is not as obvious that this is a `MaterialP` node.
Also can you show a fleshed out example here?

source/parameters/material.md
```diff
@@ -0,0 +1,42 @@
+(s:material.params)=
+## MaterialP:  Definition of materials
+
+```{code} yaml
+MaterialP:
+- state:             # ["solid", "liquid", "gas"]
+- material:          # [CompoundP, string] definition of the material in terms of compounds or named elements.
+```
+
+Materials can be accessed primarily by their name. Similar to the openPMD 2.0.0 specification, pure elements names are their symbols of the periodic table, specific isotopes are denoted by a pound symbol `#` followed by the isotopic number followed by an element, e.g.: `#3He` for Helium-3. Pure elements have their natural isotope ratios.
```
DavidSagan, line 10: When you say "Similar to the openPMD" you should indicate where there is a difference.

source/parameters/material.md
```diff
@@ -0,0 +1,42 @@
...
+
+Mixtures of multiple elements (or isotopes) in specific ratios can be accessed by their name according to Geant4's specifications [G4 Manual](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/html/Appendix/materialNames.html) or defined with a custom `CompoundP` object:
+
+
+```{code} yaml
+CompoundP:
```
DavidSagan, line 16: It is not clear from this documentation that `CompoundP` is needed and cannot be absorbed into `MaterialP` Can you explain?

source/parameters/matter.md
DavidSagan, line 2: Can you motivate why there is another layer rather than merging this with `MaterialP`?

PR #98, opened 2025-10-28
Finish `Marker`

### LINKED ISSUE #97: Finishing Marker
The `Marker` is still `Under construction ...`.

We should finalize it, remove the note, and maybe consider giving it no parameter groups?

Image

ax3l:
Finish the `Marker` element kind.

Close #97

### DIFF
--- a/source/element-kinds.md
+++ b/source/element-kinds.md
@@ -313,8 +313,7 @@ Element parameter groups associated with this element kind are:
 ## Marker Element
 
 Zero length element to mark a particular position.
-
-Under Construction...
+The main purpose of this element is to name a position in the beamline.
 
 Element parameter groups associated with this element kind are:
 - [**ApertureP**](#s:aperture.params): Aperture parameters.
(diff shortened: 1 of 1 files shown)


source/element-kinds.md
```diff
@@ -313,8 +313,7 @@ Element parameter groups associated with this element kind are:
 ## Marker Element
 
 Zero length element to mark a particular position.
-
-Under Construction...
+The main purpose of this element is to name a position in the beamline.
 
 Element parameter groups associated with this element kind are:
```
ax3l: Do we really want these parameter groups on `Marker`?

cemitch99 (approved):

source/element-kinds.md
DavidSagan: The idea is that a `Marker` element has a unit transport map (that is, does not affect the particle phase space coordinates). So it is not "generic" in the sense that it could represent a zero length kicker. However, for example, using markers for BPMs would be OK.

I don't know what is meant by "Do we really want these parameter groups on Marker?"

DavidSagan: Yes I agree. I would take out `ReferenceChangeP` as a possible group for `Marker` elements.

Issue #86, opened 2025-10-22
Renaming a few Elements


ax3l:
When working on the Python implementation, I noticed the following elements have inconsistent/unnecessarily abbreviated/superfluous suffixes in their names.

I would suggest the following renamings.

- `BeginningEle`: -> `Begin` or `Start`,
   (or maybe even a generic `Source` of particles, as in ImpactX)
- `NullEle`: -> `Empty`
  (note: the names `Null` and `None` are reserved keywords in some of the file formats and languages that PALS supports, so we should avoid them)
- `UnionEle`: -> `Union` or `Overlap`

ax3l:
Got it. I think if we use the `Ele` suffix then writing non-abbreviated `Element` could be more readable.

In practice, the elements will always be used in the context of a beamline/PALS definition as elements and in codes are likely in some kind of `element` namespace. Thus, I do feel the `Element` suffix does not add too much meaning, which is why we do not append it to other elements either.

DavidSagan:
Yes `Element` would be more readable but I do think `Ele` would be good enough. Let us see what others think...

jsberg-bnl:
I started thinking about this more as the discussion went on, and I'm wondering whether doing its functionality as an element is really the right thing. Here's where I'm coming from: the purpose of the beginning element is to set initial reference parameters (energy, time, maybe a full orbit) and layout parameters (floor position). None of these is really a property of an element per se, but is a propagated property when the element is put into a branch. So shouldn't the functionality that we're trying to assign to the beginning element really be assigned to the branch? Basically "here are the reference parameters at the start of the branch" or "pull the initial reference parameters for the start of the branch from element blah."

DavidSagan:
Things get complicated when stuff like Twiss parameters, orbit, or floor position are to be specified at someplace that is not the beginning of the branch. And needed is being able to designate multiple such places and to be able to switch from one to the other as the "active" element. So here is a proposal: Rename `BeginningEle` to something like `FixerEle` (since it can be used to "fix" Twiss and/or other parameters at a particular location).

PR #101, opened 2025-10-29
Rename: `NullEle` -> `Placeholder` Element

ax3l:
Removed `Ele` suffix, as I think [the sporadically used `Ele` suffix is repetitive, abbreviated/unclear, inconsistent, and simply not needed _in practice_.](https://github.com/pals-project/pals/pull/101#issuecomment-3469230396).

Renamed from `Null` because `Null`/`null`/`None`/`none` are reserved keywords in the file formats and programming languages we support in PALS. Related to #86

## Votes

- [x] [No suffix: adopted](https://github.com/pals-project/pals/pull/101#issuecomment-3492153377)
- [x] [New name of this element](https://github.com/pals-project/pals/pull/101#issuecomment-3492265097)

### DIFF
--- a/source/element-kinds.md
+++ b/source/element-kinds.md
@@ -386,18 +386,18 @@ Element parameter groups associated with this element kind are:
 - [**TrackingP**](#s:tracking.params): Tracking parameters.
 
 %---------------------------------------------------------------------------------------------------
-(s:empy)=
-## Empty Element
+(s:placeholder)=
+## Placeholder Element
 
 Placeholder element used for bookkeeping when constructing the expanded lattice.
-This element is empty, zero length and does nothing.
+This element has zero length and does nothing.
 This element can be used as a [`base_item`](#s:placement) element for [superpositions](#s:superposition). 
 Additionally, this element can be used, for example, to denote an invalid element in the internal
 structures defined by a program.
 
-`Empty` elements present in a lattice file will, as a part of lattice expansion, be removed.
-That is, `Empty` elements will never be present in the final expanded lattice and
-tracking through a `Empty` will never be needed.
+`Placeholder` elements present in a lattice file will, as a part of lattice expansion, be removed.
+That is, `Placeholder` elements will never be present in the final expanded lattice and
+tracking through a `Placeholder` will never be needed.
 
 This element does not have any associated parameter groups.
 


EZoni (approved):

DavidSagan:
I think `Empty` is too generic. Better is `EmptyEle`.

ax3l:
I think the spurious `Ele` suffix is repetitive, abbreviated/unclear, inconsistent, and simply not needed _in practice_.

Element kinds do not float as individual words around in PALS, they are only defined in a _context_.

## PALS Files that Users Exchange (YAML/JSON/etc.)

In PALS files the elements will always appear after an _element kind_, unmistakenly clear.
```yaml
my_drift:
  kind: Drift
do_nothing_element:
  kind: Empty
```

## APIs

In language implementations, people will create name spaces, e.g., Python
```py
pals.kinds.Drift(...)
pals.kinds.Empty(...)
```
or C++:
```py
pals::kinds::Drift(...);
pals::kinds::Empty(...);
```

It is really not a problem once you look at in which _context_ this will be used in practice.

DavidSagan:
To my mind `EmptyEle` is much clearer than `Empty`. In any case, there should be a poll to see what people think.

DavidSagan:
I like "null" over "empty" since "null" emphasizes that this element, in one sense, does not exist and is only being used as a placeholder.

ax3l:
Yep, I generally like "null" (and "none") too, but it will collide with keywords reserved in `yaml`/`json` as well as programming languages...

DavidSagan:
`NullEle` is not a reserved word.

DavidSagan:
It is not spurious nor unclear (it is much clearer than using `Null` or the like). If you don't like `NullEle` how about `NullElement`? Also it is not a question of whether it is needed. That was never the issue. The issue is does it make things clearer for Users in general? I would argue that for Users who are relatively new to PALS, this does make things clearer.

DanielWinklehner:
I personally like "Stub"

But...
What about "Nihil" (Latin for "nothing"). That shouldn't interfere with existing programming syntax.
What about "Placeholder?" It's too long for my taste, but not longer than NullElement.

mattsignorelli:
If it is always in the context Axel describes in https://github.com/pals-project/pals/pull/101#issuecomment-3469230396,  where it is never unclear that `Empty` refers to an empty element, then I would agree with Axel that this is the cleaner more consistent solution.

DanielWinklehner (approved):

DavidSagan:
@EZoni No.

EZoni:
@ax3l 

There's a tie between `kind: Empty` and `kind: Placeholder` - which one do you want to pick? I'm fine with the PR as is.

EZoni:
It's a tie again - `Placeholder` is not in the lead anymore.

ax3l:
We discussed this quickly in the @pals-project/pals-technical-committee today and since this element is doing nothing, there is no big harm of leaving it in for now. It is a literal NO-op.

We can revisit if really nobody uses it in the future and remove it if so.

PR #109, opened 2025-10-30
Graphic change: `Lattice File` -> `PALS Lattice File`.

DavidSagan:


source/figures/translator.svg
EZoni: Note that the box around `PALS Lattice File (JSON, YAML, ...)` seems too short now - the `...)` is falling out of it.

DavidSagan:
@EZoni I changed the line. Note that "lattice file" always refers to a PALS lattice file except where explicitly noted. In general, when the documentation for any particular program (Bmad, ELEGANT, etc.) refers to a "lattice file" it is understood to mean a lattice file for that particular program.

PR #112, opened 2025-11-05
Added misalignment figures.

DavidSagan:


cemitch99 (approved):

ax3l (approved):
At least I see them [rendered in the preview](https://pals-project--112.org.readthedocs.build/en/112/element-parameters.html#bodyshiftp-element-body-orientation-and-position-parameters), which is good enough :D

PR #120, opened 2025-11-11
Add sextupole, octupole, multipole, and solenoid.

cemitch99:
This short PR adds the following basic element kinds:

- Sextupole
- Octupole
- Multipole
- Solenoid

### DIFF
--- a/source/element-kinds.md
+++ b/source/element-kinds.md
@@ -528,7 +528,7 @@ s01w:
 (s:solenoid)=
 ## Solenoid Element
 
-A `solenoid` is an element whose magnetic field is dominated by a uniform field whose direction is aligned with the magnetic axis.
+A `solenoid` is an element whose magnetic field is dominated by a field whose direction is aligned with the magnetic axis.
 Additional magnetic (or electric) multipole contributions are allowed.
 
 Element parameter groups associated with this element kind are:


source/element-kinds.md
```diff
@@ -509,11 +517,19 @@ Element parameter groups associated with this element kind are:
...
+  kind: Sextupole
+  length: 0.5
+  Kn2: 0.28
+```
 %---------------------------------------------------------------------------------------------------
 (s:solenoid)=
 ## Solenoid Element
 
-Under Construction...
+A `solenoid` is an element whose magnetic field is dominated by a uniform field whose direction is aligned with the magnetic axis.
```
DavidSagan: I would remove the word `uniform` since there are cases where the field is not.

source/element-kinds.md
```diff
@@ -407,9 +406,9 @@ For other purposes, for example, to mark reference points, [Marker](#s.marker) e
 (s:octupole)=
 ## Octupole Element
 
-Octupole element.
-
-Under Construction...
+An `octupole` is an element whose major field has a cubic field dependence with transverse offset.
+Both electric and magnetic fields can be defined and additional multipole contributions are allowed.
+In terms of functionality, an `octupole` is equivalent to a [`Multipole`](#s:multipole) element.
```
EZoni: What does "major field" mean in this context? Does it imply any constraint on the field components that should be set up or can an octupole have, say, `KnN` components of any order `N`?

In other words, should/could we validate somehow whether an octupole is indeed an octupole? 

I would ask the same questions for sextupole, etc.

EZoni (approved):

PR #122, opened 2025-11-11
Rename conda environment, update documentation

EZoni:
I propose that we rename the conda environment `pals`. The current name `lattice-standard` was matching the original name of the project. I also propose an update to the documentation on how to write documentation.


DavidSagan:
Since this environment can be used for any Myst documentation (and I am using this in another repo), why not rename the environment `myst`?

ax3l:
We might modify it further later, e.g., importing pals-specific packages/dependencies.

cemitch99 (approved):

PR #135, opened 2025-12-09
Conventions Updates & Matching Syntax

DavidSagan:
Conventions modifications moved from PR #119.

### DIFF
--- a/source/conventions.md
+++ b/source/conventions.md
@@ -348,7 +348,7 @@ asinh(x), acosh(x)       # Hyperbolic arc sine and Arc cosine
 atanh(x), acoth(x)       # Hyperbolic arc tangent and cotangent
 abs(x)                   # Absolute Value
 factorial(n)             # Factorial
-ran()                    # Random number between 0 and 1
+random()                 # Uniform random number between 0 and 1
 ran_gauss()              # Gaussian distributed random number
 ran_gauss(sig_cut)       # Gaussian distributed random number
 int(x)                   # Nearest integer with magnitude less then x


source/conventions.md
```diff
@@ -220,3 +278,59 @@ The lattice standard uses SI except for energy which uses `eV`.
...
+r_p                       # [m] Classical proton radius
+e_charge                  # [Coul] Elementary charge
+mu_0_vac                  # [eV*sec^2/m] Vacuum permeability
+eps_0_vac                 # [1/eV*m] Permittivity of free space
+classical_radius_factor   # [m*eV] Classical Radius Factor: 1/(4 pi epsilon_0 c^2)
+fine_structure_const      # [-] Fine structure constant
+n_avogadro                # [-] Avogadro's constant
+```
+The `classical_radius_factor` is a useful number when converting a formula that involve the classical
+electron or proton radius to a formula for something other than an electron or proton.
```
EZoni: We had this debate a few times for some of the BLAST codes in the past and we may want to have it here too.

I think many scientists are familiar, and used to, SciPy's naming conventions for constants, defined in https://docs.scipy.org/doc/scipy/reference/constants.html.

Maybe it could be beneficial to follow the same naming conventions (e.g., `epsilon_0` instead of `eps_0_vac`, `hbar` instead of `h_bar_planck`, etc.) as much as possible. 

I would personally be in favor of this, but I suggest that we could vote on it.

source/conventions.md
```diff
@@ -171,23 +110,140 @@ lattice expansion where there are [`Fork`](#s:forking) elements and for evaluati
...
+(henceforth just referred to as a branch name) to the string.
+using the string `">>"` as a separator. For example, `"B1>>Sextupole::Saf"` would match
+to all Sextupole elements in a branch or BeamLine named `"B1"` whose name was `"Saf"`.
+This includes sublines of BeamLines. Thus if `B1` is a BeamLine that contains a subline `B2` 
+that in turn contains a Sextupole element named `Saf`, the string `"B1>>Sextupole::Saf"`
+will match to this element.
 
 Lattice elements can also be referred to by the index in which they appear in a branch
-with the first element having index one, etc. Branches do not get an index since the
+with the first element having index one, etc. For example, `"B1>>7"` matches the seventh
```
ax3l: The other item we want to discuss in the meeting: shall PALS follow zero-based (C/C++/Python/etc.) or one-based (Fortran/Julia) indexing for arrays/lists?

source/conventions.md
```diff
@@ -171,23 +110,140 @@ lattice expansion where there are [`Fork`](#s:forking) elements and for evaluati
...
+where `<kind>` is the element kind and `<name>` is the element name.
+Example:
+```{code} yaml
+Marker::bpm.
+```
+This will match to all `Marker` elements whose name is four characters starting with `bpm`.
+
+The `N`{sup}`th` element with a given name can be matched to by appending the character `"#"` 
+followed by an integer `N`. For example, `"Quadrupole::Q1#3"` will match to the third element
+that matches `"Quadrupole::Q1"`. Notice that the `N`{sup}`th` instance selection is applied last.
```
EZoni: What does "Notice that the `N`-th instance selection is applied last." mean exactly? Could you give an example?

source/conventions.md
```diff
@@ -171,23 +110,140 @@ lattice expansion where there are [`Fork`](#s:forking) elements and for evaluati
...
+<!--
+Besides the element name, any parameter in the [`MetaP`](#s:meta.params) parameter group can
+be matched to using the syntax `"{param-name}::{name-to-match}"`. For example, `"alias::black"`
+would match to any element whose `alias` parameter is set to `black`.
+-->
+
+In the discussion below, all of the above constructs will be called an "element name".
+
+The names of an element may be "qualified" by prepending a `branch` or `BeamLine` name 
+(henceforth just referred to as a branch name) to the string.
```
EZoni:
```suggestion
(henceforth just referred to as a branch name) to the string,
```

source/conventions.md
```diff
@@ -171,23 +110,140 @@ lattice expansion where there are [`Fork`](#s:forking) elements and for evaluati
...
+using the string `">>"` as a separator. For example, `"B1>>Sextupole::Saf"` would match
+to all Sextupole elements in a branch or BeamLine named `"B1"` whose name was `"Saf"`.
```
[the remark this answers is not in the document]
EZoni: Maybe I lost track of which convention we follow when we refer to specific element kinds or parameters in the text. I thought we wrap them in \`...\` to produce `...` (e.g., we write `BeamLine` (i.e., not BeamLine) if we use the exact name and simply beam line if we refer to the concept), but maybe we don't do that anymore (or never did). Just checking to make sure we remain consistent.

source/conventions.md
EZoni: Seems like 1-based indexing was preferred, correct?

Image

Not sure what conclusion to infer from the PCRE2 and CSS comments.

source/conventions.md
```diff
@@ -171,23 +110,140 @@ lattice expansion where there are [`Fork`](#s:forking) elements and for evaluati
...
+Element matches may be restricted to a given element kind using the notation
+```{code} yaml
+<kind>::<name>
+```
+where `<kind>` is the element kind and `<name>` is the element name.
+Example:
+```{code} yaml
+Marker::bpm.
+```
+This will match to all `Marker` elements whose name is four characters starting with `bpm`.
```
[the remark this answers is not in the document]
EZoni: I just saw that there is some more info on this when we talk about regular expressions below, maybe we could refer to that paragraph here.

source/conventions.md
```diff
@@ -171,23 +110,140 @@ lattice expansion where there are [`Fork`](#s:forking) elements and for evaluati
...
+Order of presidence:
+```{code} yaml
+>>>     # Highest
+>>
+::      
+#
+:
+,
+&      # Lowest
+```
```
EZoni: General question: Is this arbitrary or did it come from somewhere? I'm asking about both the characters and their order.

source/conventions.md
DavidSagan: There is nothing in the Xsuite conventions that I would use for PALS. Note: with PALS there are functions for mass. For example: `mass_of(electron)`.

DavidSagan: I'm not sure how many people are familiar with the SciPy constants. In any case, I am not in favor of using short names like `c` and `h` for constants since this could be confusing. But we can have a vote on this if someone wants it. @EZoni Do you want to make specific suggestions now? Alternatively, we can use what is here as a starting point and future PRs can be made when changes/additions are wanted. What do you think?

source/conventions.md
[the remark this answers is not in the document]
DavidSagan: Changed in next push (done shortly). I added wording for this.

source/conventions.md
EZoni: @DavidSagan 

I'm fine with not blocking the PR per se, but I will anyways open a few targeted GitHub polls for the ones that I would like to suggest.

DavidSagan:
> @ax3l @DavidSagan
> 
> If you want, I could open a [Poll](https://github.com/orgs/pals-project/discussions/categories/polls) to let people vote on the following options:
> 
> 1. `UPPER_CASE` always (e.g., `EPSILON_0`, `N_AVOGADRO`, etc.)
> 2. `lower_case` always (e.g., `epsilon_0`, `n_avogadro`, etc.)
> 3. `mixed_Case` as needed (e.g., `epsilon_0`, `n_Avogadro`, etc.)
> 
> I can also remove option 1. since we already discarded it in the live poll.
> 
> Feel free to edit this comment if you want to refine the way we phrase the options.

If we go to mixed case I would remove the underscores and capitalize the first word. EG: `NAvogado`.

source/conventions.md
DavidSagan: Changed in next push (done shortly). I added wording to make this clear.

source/conventions.md
DavidSagan: Adapted from Bmad.

source/conventions.md
```diff
@@ -220,3 +278,59 @@ The lattice standard uses SI except for energy which uses `eV`.
...
+int(x)                   # Nearest integer with magnitude less then x
+nint(x)                  # Nearest integer to x
+sign(x)                  # 1 if x positive, -1 if negative, 0 if zero
+floor(x)                 # Nearest integer less than x
+ceiling(x)               # Nearest integer greater than x
+modulo(a, p)             # a - floor(a/p) * p. Will be in range [0, p).
+mass_of(A)               # Mass of particle A
+charge_of(A)             # Charge, in units of the elementary charge, of particle A 
+anomalous_moment_of(A)   # Anomalous magnetic moment of particle A
+```
```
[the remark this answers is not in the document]
DavidSagan: Personally I don't like `ceil`. Changing to `random` is fine. Do you want to discuss changes now or merge and discuss changes in future PRs?

source/conventions.md
EZoni: FYI, I opened a few initial polls: #140, #141, #142, #143, #144, #145.

DavidSagan:
@ax3l I do not believe there was a consensus. In any case, the voting on the names will establish what people want.

source/conventions.md
DavidSagan: I have updated the PR to reflect the polls.

PR #158, opened 2026-01-28
Lattice/Fork: Element Syntax

ax3l:
Update the lattice/fork code snippets to use the same schema we use for elements:
```yaml
- my_lattice:
    kind: Lattice
    branches:
      - ...
```

### DIFF
--- a/source/lattice-construction.md
+++ b/source/lattice-construction.md
@@ -199,7 +199,7 @@ is the last lattice. This default can be overridden by a `use` statement. Exampl
         ...
 
 - lat2:
-    - name: Lattice
+    - kind: Lattice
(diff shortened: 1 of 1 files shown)


source/lattice-construction.md
```diff
@@ -194,13 +193,13 @@ the reference properties at the `Fork` element.
...
-    - name: lat1
+- lat1:
+    - kind: Lattice
     - branches:
         ...
 
-- Lattice
-    - name: lat2
+- lat2:
+    - kind: Lattice
```
ax3l: Ah these two examples have one more bug in them. Fix in #165

PR #159, opened 2026-02-02
Aperture: Element Syntax

ax3l:
Update another aperture code snippets to use the same schema we use for elements.

Same as #158, outdated example from a new PR.

### DIFF
--- a/source/conventions.md
+++ b/source/conventions.md
@@ -107,9 +107,9 @@ Non-YAML syntax used here is:
 1. The {math}`N^{th}` item in a list is referred to using square brackets enclosing the index: `[N]`.
 For example:
 ```{code} YAML
-Aperture:
-  name: ap1
-  x_limit: [-0.03, 0.04]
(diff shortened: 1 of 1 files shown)


source/conventions.md
```diff
@@ -107,9 +107,9 @@ Non-YAML syntax used here is:
...
 For example:
 ```{code} YAML
-Aperture:
-  name: ap1
-  x_limit: [-0.03, 0.04]
+- ap1:
+    kind: Aperture
+    x_limit: [-0.03, 0.04]
 ```
 here `x_limit[1]` and `x_limit[2]` would refer to the first and second values of `x_limit` respectively.
```
ax3l: Actually the whole example can be removed, this is a leftover from a @DavidSagan PR where we decided for a simpler syntax in the end, no?

ax3l:
Will merge because cosmetic, then merge to #138

PR #183, opened 2026-02-19
Semantic Lattice `load`

ax3l:
This improves the semantic include feature from #152 #177: This allows to compose complex lattices over multiple PALS files, which otherwise are fully standalone and standard compliant files.

This is useful to model only a sub-system at a time but still compose complex facilities. The `load` feature inside a lattice thus complements our verbatim `include` feature (which in turn is for verbatium snippets).

Summary:
- add new `load` attribute to `Lattice` kind and examples
- move the `Use statement` section slightly up and improve its example, because the `use` needs to be explained earlier now

### DIFF
--- a/source/lattice-construction.md
+++ b/source/lattice-construction.md
@@ -5,11 +5,12 @@
 (s:lattice.construct)=
 ## Constructing a Lattice
 
-A `Lattice` contains a set of branches. The component of `Lattice` is:
+A `Lattice` contains a set of branches. The components of `Lattice` are:
 ```{code} yaml
-branches        # [List] List of branches.
+branches        # [List] List of branches
+load            # [List of paths] Another PALS file that defines a lattice
 ```
-Each branch is instantiated from a `BeamLine` that is called the `root BeamLine`.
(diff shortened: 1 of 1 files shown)


source/lattice-construction.md
```diff
@@ -35,6 +36,78 @@ periodic  # [Boolean] Optional. Are orbit and Twiss parameters periodic?
...
+```{code} yaml
+# Content of storage_ring.pals.yaml
+PALS:
+  version: ...
+  facility:
+    - storage_ring:
+        kind: Lattice
+        load:
+          - "injector.pals.yaml"
+          - "accumulator.pals.yaml"
```
DavidSagan: What happens if a loaded file has multiple lattices?

EZoni:
@ax3l 

I think @DavidSagan left a few comments that need to be addressed before finalizing this PR.

DavidSagan:
NOTE: #260 if accepted will supersede this PR.

ax3l:
Continued in #260 :)

PR #188, opened 2026-03-02
File name suffix convention is recommended but not required.

DavidSagan:
It is highly unusual for file name suffixes to be mandated. PALS should not be an exception and the choice should be left up to the User.


cemitch99 (dismissed):

ax3l (approved):
That looks good to me, thank you.

cemitch99 (approved):

PR #203, opened 2026-03-09
Proposal: An Ordered list w/ unique keys can be used in place of a unordered dictionary

DavidSagan:
Clarified that ordered list with unique keys can be used in place of a unordered dictionary.

Rationale: A PALS parser may not always be able to preserve insertion order when it reads in a dict. And it is sometimes desirable to preserve the order. In particular, the [C++ parser being developed](https://github.com/pals-project/pals-cpp) does not have this capability since it uses an external parsing library that does not have this feature. (*)

Correction added by @ax3l: that premise (*) is wrong. The currently picked library (and many other C++ libs for YAML, JSON, etc.) do in fact support preservation of insertion order. This is a [bug in pals-cpp](https://github.com/pals-project/pals-cpp/issues/18) (and [here is the fix](https://github.com/pals-project/pals-cpp/pull/19)); it is not a library issue and not a fundamental limitation.

### DIFF
--- a/source/conventions.md
+++ b/source/conventions.md
@@ -53,8 +53,11 @@ this_list:
       key4: value4
 ```
 
+In a lattice file, an ordered list of key-value pairs can always be used in place of a dictionary
+with the restriction that no duplicate keys can be present in the list.
+
 ```{note}
-   Developer note:
+   Code Developer note:
    PALS dictionaries should, when possible, implement a dictionary that preserves insertion order.
 
    While not strictly necessary, this helps with human readability:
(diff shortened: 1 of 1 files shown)


cemitch99 (dismissed):

EZoni (dismissed):

ax3l (changes requested):
Review summary for status on the PR: I oppose this, this is not a good idea.

Details described here: https://github.com/pals-project/pals/pull/203#discussion_r3017943306

DavidSagan:
> The best validation check is the one _you_ don't need to write.

I agree. But I believe this is being blown all out of proportion.

jlvay (commented):
After finally understanding what this PR was doing and reviewing, I think that the proposed option would bring confusion and additional work for the parsers that is not needed.

jlvay:
Decided to close after extensive discussions.

PR #209, opened 2026-04-02
YAML 101: Show JSON, TOML, Python, C++ & Julia

ax3l:
## Motivation

From feedback from early PALS users and contributors that have not used YAML yet much, the YAML 101 can be slightly improved with better examples: By showing equivalent examples of the two fundamental container types (list and dict) in JSON, TOML, Python, C++ and Julia we can make it easier for people new to YAML to read the rest of the document.

## Compact Implementation

By using a "tab-set" (see [preview here](https://pals-project--209.org.readthedocs.build/en/209/conventions.html#syntax-used-in-this-document-is-yaml)) this adds no scroll bloat to someone reading the standard that already knows YAML.

## Alternative Placements

I also considered moving this in the end to _file formats_, but it breaks **linear reading** of the standard early on, if one has to jump there to learn the conventions used in the document, jump back to continue reading, etc. As a long-term documentation writer, I would avoid such literal and mental jumps.

One could also move [file formats](https://pals-project--209.org.readthedocs.build/en/209/fileformats.html) as a whole forward to _conventions_. Open to ideas.

### DIFF
--- a/source/conventions.md
+++ b/source/conventions.md
@@ -28,22 +28,148 @@ YAML is [formally defined](https://yaml.org) and there are [tutorials to learn Y
 ### YAML 101
 
 In particular, there are dictionaries, which are unordered sets of key-value pairs, e.g.:
+
+::::{tab-set}
+
+:::{tab-item} YAML
+:sync: yaml
 ```{code} yaml
 this_dictionary:
   key1: value1
-  key2: value2
-  key3: value3
+  key2: 2.0
+  key3: true
 ```
+:::
+:::{tab-item} JSON
+:sync: json
+```{code} json
+{"this_dictionary": {
+  "key1": "value1",
+  "key2": 2.0,
+  "key3": true
+}}
+```
+:::
+:::{tab-item} TOML
+:sync: toml
+```{code} toml
+[this_dictionary]
+key1 = "value1"
+key2 = 2.0
+key3 = true
+```
+:::
+:::{tab-item} Python
+:sync: python
+```{code} python
+{"this_dictionary": {
+  "key1": "value1",
+  "key2": 2.0,
+  "key3": True
+}}
+```
+:::
+:::{tab-item} C++17
+:sync: cpp
+```{code} c++
+// insertion-order preserving dictionary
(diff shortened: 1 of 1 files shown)


ax3l:
I think we can also expand `fileformats.md` :+1: 

Why I add it here is to make it easier for people to learn YAML 101 when they already know other formats or programming languages. The section is part of `Syntax Used in this Document is YAML` that already mentions other languages/formats and links to `fileformats.md`, so it is not out of the blue to show the example.

Since it uses tabs, it does not add bloat / scroll burden when reading the short section.

jlvay (approved):
Looks good to me and I think that it is super useful.

jlvay:
> Yes, I also considered moving this to the end in _file formats_, but it breaks **linear reading** of the standard early on, if one has to jump there to learn the conventions used in the document, jump back to continue reading, etc. As a long-term documentation writer, I would avoid such early literal and mental jumps.
> 
> One could also move the small section [file formats](https://pals-project--209.org.readthedocs.build/en/209/fileformats.html) as a whole forward to _conventions_. Open to ideas.

I agree with the placement of the YAML 101 in the conventions given that YAML is used for the examples all over the documentation.

source/conventions.md
```diff
@@ -28,22 +28,148 @@ YAML is [formally defined](https://yaml.org) and there are [tutorials to learn Y
...
+
+::::{tab-set}
+
+:::{tab-item} YAML
+:sync: yaml
 ```{code} yaml
 this_list:
   - entry1
   - entry2
   - more_entries
```
EZoni: We could also rename `more_entries` as `entry3`, for consistency with the dictionary example, which has three keys named `key1`, `key2`, and `key3`. Happy to make the change if you agree but don't want to spend more time on this. Just a suggestion.

EZoni (changes requested):

EZoni (approved):

ax3l:
This seems generally a welcome addition.

I will propose another PR where we move file format introductions forward and split the now a bit vague "conventions" section into a bit clearer parts.

PR #214, opened 2026-04-21
Revamped reference energy/time/species adjustments.

DavidSagan:
Rational: Reference adjustments should not be part of "regular" elements (like quadrupoles, etc.) since the fact that such an adjustment is being used is obscured. With the current PR, adjustments are confined to be in a specific `ReferenceChange` element.

### DIFF
--- a/source/element-kinds.md
+++ b/source/element-kinds.md
@@ -213,21 +213,6 @@ q01w:
 ```
 
 %---------------------------------------------------------------------------------------------------
-(s:reference.change)=
-### ReferenceChange Element
-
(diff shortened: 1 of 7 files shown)


cemitch99:
Should the `ReferenceChange` element kind be placed in the section containing the "Bookkeeping" elements, instead of the "Magnets and RF Cavities" section?  The best category is not 100% clear.

DavidSagan:
@cemitch99 Yes moved to `Bookkeeping`.

cemitch99 (approved):

PR #219, opened 2026-05-06
Minor correction for displayed equations with underscores.

DavidSagan:


EZoni:
There are currently [issues](https://www.githubstatus.com/) with creating new PR threads, e.g., line and file comments, so I cannot suggest changes properly, but here's one thing I noticed.

One of your changes displays as

Image

Was this the intention, i.e., to display `x_rot` and `y_rot`, or did you want to display $x_{\text{rot}}$ and $y_{\text{rot}}$? For the latter, I think it should be 
```
  \cos(x_{\text{rot}}) \, \cos(y_{\text{rot}})
```
or probably even simply (to double check)
```
  \cos(x_\text{rot}) \, \cos(y_\text{rot})
```

DavidSagan:
@EZoni I am using the parameter name `x_rot` rather than inventing a corresponding math symbol. If you think using a corresponding math symbol is better, I will change. I don't much care which is used.

EZoni:
I'm fine with either as well, I just wanted to double check that this had not been overlooked. If `x_rot` and `y_rot` match PALS names, I guess it's fine to use them directly in math formulas.

EZoni (approved):

PR #220, opened 2026-05-06
Remove `Lattice Standard` in favior of `Language Standard` or `PALS`.

DavidSagan:
This PR is essentially to remove instances of `Lattice Standard`. More work needs to be done in modifying the text in the README.md file and elsewhere but this is left to later PRs.

### DIFF
--- a/source/code-packages.md
+++ b/source/code-packages.md
@@ -21,7 +21,16 @@ uses the C++ parser as a back end.
 ## Visualization
 
 A 3D Visualization package is in development.
-Xelera Research LLC is working to develop an open-source extension to the widely used graphics program Blender that will provide an intuitive graphical user-interface for accelerator modeling in 3D. The extension will support lattice editing in both a 2D graphical node-based layout as well as in the 3D view, and it will include support for importing and exporting the PALS lattice standard format. The goal is for this software to enable viewing, editing, or constructing beamline models with a 3D layout that can include realistic physical constraints and infrastructure. Xelera also plans to explore incorporating external signals into the software to permit integration with control systems at accelerator facilities. This would enable such signals to be coupled with a model to permit continuously updated output in the 3D view. This work is supported by the U.S. Department of Energy SBIR program.
+Xelera Research LLC is working to develop an open-source extension to the widely used graphics 
+program Blender that will provide an intuitive graphical user-interface for accelerator modeling in 3D. 
+The extension will support lattice editing in both a 2D graphical node-based layout as well as in 
+the 3D view, and it will include support for importing and exporting the PALS standard format. 
+The goal is for this software to enable viewing, editing, or constructing beamline models with a 3D 
+layout that can include realistic physical constraints and infrastructure. 
+Xelera also plans to explore incorporating external signals into the software to permit integration 
+with control systems at accelerator facilities. 
+This would enable such signals to be coupled with a model to permit continuously updated output in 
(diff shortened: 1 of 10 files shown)


source/coordinates.md
EZoni, line 1: I think the changes to source/coordinates.md are not related to this PR, but rather #219, no? In fact, I'm puzzled by the fact that they show up in the diff here, given that #219 has been already merged...

EZoni: On the other hand, this instance of "lattice standard" in source/coordinates.md should probably be updated:
https://github.com/pals-project/pals/blob/a6010c0d742294947b2a979c4ea759591c428cc7/source/coordinates.md?plain=1#L388

DavidSagan: This line has been corrected. I crossed two branches which is why coordinates.md looks funny but this will not affect the final version once this PR is merged.

source/simulations.md
```diff
@@ -1,7 +1,7 @@
 (c:impl.sims)=
 # Simulations
 
-The Particle Accelerator Lattice Standard (PALS) will be used used in various accelerator lattice simulations.
+The PALS will be used used in various accelerator lattice simulations.
```
EZoni: Maybe the initial "The" here was left over? Generally, should one say "The PALS does something" (because "S" stands for standard and one would say "The standard does something") or "PALS does something"? "PALS does something" sounds more correct to me, but I'm not a native speaker. Also, `used` is repeated twice.
```suggestion
PALS will be used in various accelerator lattice simulations.
```

source/code-packages.md
```diff
@@ -21,7 +21,16 @@ uses the C++ parser as a back end.
...
+Xelera Research LLC is working to develop an open-source extension to the widely used graphics 
+program Blender that will provide an intuitive graphical user-interface for accelerator modeling in 3D. 
+The extension will support lattice editing in both a 2D graphical node-based layout as well as in 
+the 3D view, and it will include support for importing and exporting the PALS standard format. 
+The goal is for this software to enable viewing, editing, or constructing beamline models with a 3D 
+layout that can include realistic physical constraints and infrastructure. 
+Xelera also plans to explore incorporating external signals into the software to permit integration 
+with control systems at accelerator facilities. 
+This would enable such signals to be coupled with a model to permit continuously updated output in 
+the 3D view. This work is supported by the U.S. Department of Energy SBIR program.
```
EZoni: For future changes, keeping the line breaks consistent with the original version makes diffs like this easier to review - in this case, it would surface only the single substantive change. That said, I do agree that breaking long lines is generally preferable; a good convention that we could try to follow going forward is to insert a line break at the end of each sentence.

DavidSagan: Agreed.

EZoni (dismissed):

EZoni (dismissed):

EZoni:
@DavidSagan 

I will push a trivial commit to see if I can fix the merge-base issue that keeps appearing, which is probably related to my earlier comment https://github.com/pals-project/pals/pull/220#discussion_r3197976611.

EZoni:
@DavidSagan 

The wrong diff is still present, I will try to close and re-open the PR to see if it gets fixed.

EZoni (approved):

EZoni:
@DavidSagan 

I pushed some commits to minimize the diff with the base branch, please re-review the changes and let me know if they look good to you.

DavidSagan:
Looks good.

PR #223, opened 2026-06-11
Bend Parameters Revamp

DavidSagan:
Revamping of the `BendP` parameter group to include the possibility of defining a "rectangular" bend geometry.
Also bend parameters where reorganized to be more logical.
The changes are substantial enough so that my suggestion is reviewers should look at the new `BendP` doc in its entirety rather than trying to look at the differences.


qianglbl (commented):
Not sure about the following line: "One common mistake when creating orbit bumps using a bend is to vary g_ref`." Is it a mistake or just a different way to handle that?

DavidSagan:
@qianglbl It is indeed a mistake the sentence before explanes why:

> It is important to keep in mind that changing g_ref will change the branch reference orbit and hence will move all downstream lattice elements in space.

Does the paragraph need to be reworded to make things clearer?

qianglbl:
@DavidSagan It could make it clearer with a bit of rewording.

DavidSagan:
Can you suggest something?

DavidSagan:
@qianglbl OK I modified the paragraph. Please take a look...

qianglbl:
@DavidSagan It looks better. Thanks.

mattsignorelli:
I would suggest having `g_ref` be the sole definer of the coordinate system curvature / integration path, completely separate from the magnetic field strength. Connecting the two introduces problems such as defining electric dipole bends, which would then require the less-than-elegant solution of having g_ref=-g_actual or something like that.

For context, I have already heard some criticisms about this from people at BNL playing around with Bmad, which has this.

DavidSagan:
@mattsignorelli Can you construct a separate PR for that? I don't want to pile too many things into this one.

mattsignorelli:
Unfortunately I don't have the time to construct a whole PR for it, I am just expressing my thoughts and thoughts of colleagues about bend parameters I see being merged in this PR

DavidSagan:
How `g_ref` is handled is not being changed in this PR. This is a different matter. And I am open to your suggestion. I just do not want to couple it to this PR.

source/parameters/bend.md
```diff
@@ -7,86 +7,127 @@ length `length` parameter.
...
-  ```{code} yaml
-  B-field(total) = bend_field_ref + Bn0
+
+  The `bend_field_ref` parameter is the reference magnetic bending field which is the field
+that is needed for the reference particle to be bent in a circle of radius `rho_ref`.
+The direction of the reference bend field along the {math}`y`-axis.
+  ```{math}
+  :label: bff
+
+  \text{bend_field_ref} = \frac{p_0}{q} \cdot \text{g_ref}
```
cemitch99: The equation above does not appear to render properly.

cemitch99:
@jsberg-bnl expressed interest during 6/17 meeting and likely has comments -- please feel free to take a look.

source/parameters/bend.md
DavidSagan: Fixed.

cemitch99 (approved):

PR #226, opened 2026-06-12
Expanded constant definition section.

DavidSagan:
Expanded constant definition section.

### DIFF
--- a/source/miscellaneous.md
+++ b/source/miscellaneous.md
@@ -38,5 +38,5 @@ For sets that only have a value, an alternative compact form has the syntax:
 - sets:
     - param1: value1
     - param2: value2
-    ... etc...
+    ...
 ```
--- a/source/miscellaneous.md
+++ b/source/miscellaneous.md
@@ -3,13 +3,17 @@
 
 %---------------------------------------------------------------------------------------------------
 (s:set)=
-## Set command
+## Setting Parameters
 
 The `set` command is used for setting parameters. The components of `set` are:
 ```{code} yaml
-parameter     # [String] Parmeter(s) to vary.
-value         # [Expression] Value to set.
+parameter               # [String] Parmeter(s) to vary.
+value                   # [Expression] Value to set.
+absolute_error: 0       # Absolute error.       
+relative_error: 0       # [-] Relative error.
 ```
+If both `absolute_error` and `relative_error` are specified, 
+the true error is `absolute_error + relative_error * |value|`.
 In the `value` expression, `parameter` can be used for the current value of the parameter being changed
 and `self` can be used for the lattice element whose parameter is being changed. 
 Example:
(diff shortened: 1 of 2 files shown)


cemitch99 (dismissed):

cemitch99 (dismissed):

source/miscellaneous.md
```diff
@@ -3,13 +3,17 @@
...
+## Setting Parameters
 
 The `set` command is used for setting parameters. The components of `set` are:
 ```{code} yaml
-parameter     # [String] Parmeter(s) to vary.
-value         # [Expression] Value to set.
+parameter               # [String] Parmeter(s) to vary.
+value                   # [Expression] Value to set.
+absolute_error: 0       # Absolute error.       
+relative_error: 0       # [-] Relative error.
```
EZoni, line 13: I might not remember our latest syntax, what is the meaning of `[-]` in the comment here?

source/miscellaneous.md
```diff
@@ -28,3 +32,11 @@ In this example, the `BendP.e1` parameter of all elements whose name begins with
 This includes element `B1a`. 
 
 Note: Pattern matching is not supported in the `value` expression.
+
+For sets that only have a value, an alternative compact form has the syntax:
+```{code} yaml
+- sets:
+    - param1: value1
+    - param2: value2
+    ... etc...
```
EZoni: Same style as the examples above:

```suggestion
    ...
```

source/miscellaneous.md
DavidSagan: No units.

DavidSagan:
@EZoni Ready for rereview...

source/miscellaneous.md
EZoni: Okay. Maybe something we could change going forward, because I see we use `[]` in the lines above to denote the type (e.g., `[String]`, `[Expression]`) and also here to denote the units (or lack thereof), which could be confusing.

EZoni (approved):

source/miscellaneous.md
cemitch99: I agree with @EZoni.  Based on the context, it looks like `[-]` denotes a variable type, and it would be good to have a distinct notation for specifying units that can be consistently used.  Also, note that no units are specified for the other parameters such as `absolute_error` (because these will vary).

cemitch99: Here is a possibility:  One could end each string with variable units in parentheses, and be verbose in cases above such as:  `(variable units)` and `(no units)`.

DavidSagan: This needs discussion so best is to leave this to a future PR.

cemitch99 (approved):
Specification of units for parameters in the PALS document should appear in a follow-up PR.

DavidSagan:
> Specification of units for parameters in the PALS document should appear in a follow-up PR.

@cemitch99 Do you want to open an issue so we don't forget?

PR #227, opened 2026-06-15
Underscores in text should now be accepted.

DavidSagan:
Underscores in text should now be accepted.

### DIFF
--- a/source/conf.py
+++ b/source/conf.py
@@ -49,4 +49,5 @@ html_theme_options = {
     "show_navbar_depth": 3,
     "collapse_navbar": False,
 }
-## html_static_path = ['_static']
+html_static_path = ['_static']
+html_css_files = ['math-overflow.css']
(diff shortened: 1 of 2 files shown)


source/conf.py
```diff
@@ -28,6 +28,14 @@
...
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
 
+# Allow '_' to be used literally inside \text{...} in math (MathJax otherwise
+# raises "'_' allowed only in math mode").
+mathjax3_config = {
+    'tex': {
+        'packages': {'[+]': ['textmacros']},
+    },
+}
+
```
EZoni: So this change lets you write `_` instead of `\_` in text mode? Or was `\_` also not accepted?

DavidSagan:
Looks like this change did not fix the problem so please do not merge.

DavidSagan:
Does not solve underscore problem. Will close.

PR #231, opened 2026-06-17
Extensions: Clarify documentation.

DavidSagan:
Clarify extension documentation.


source/definitions.md
```diff
@@ -80,3 +80,12 @@ of a branch, of constructing the ordered list of lattice elements contained in t
...
 calculating the reference energy for all elements.
 
+%---------------------------------------------------------------------------------------------------
+(s:extensions)=
+## Extensions
+
+An Extension is data added to a PALS lattice that is not part of the PALS standard.
+For example, information specific to a particular machine that has no analogue with other
+machines can be put in as an extension. A list of "registered" extensions is in the 
+[Extension](#c:extension) section.
```
EZoni: I think it would be good if we clarify here what "registered" in `"registered" extensions` means. I wasn't sure myself while reviewing this part of the PR. I think the fact that we wrap it in quotes signals that it requires an explanation, which the quotes themselves don't really provide.

source/extensions.md
```diff
@@ -6,31 +6,29 @@ For example, information specific to a particular machine or simulation program
 that has no analogue with other machines or simulation programs can be put in as an extension. 
 
 Maintenance of an extension schema falls outside of PALS governance. 
-Extension maintainers can add documentation to this section by submitting a [pull request](#s:contribute).
-Documentation here should be brief and should have a reference to full documentation maintained
+Extension maintainers can add extension documentation submitting a [pull request](#s:contribute).
+If the amount of documentation is "large", It is encouraged that the documentation put in the PALS
+repo should be relatively brief and there should be a reference to full documentation maintained
```
EZoni:
```suggestion
If the amount of documentation is "large", it is encouraged that the documentation put in the PALS
repository should be relatively brief and there should be a reference to full documentation maintained
```

source/definitions.md
EZoni: Maybe we could simply write "documented" instead, as in `documented extensions`. I see we use this in other parts of this PR.

source/extensions.md
```diff
@@ -6,31 +6,29 @@ For example, information specific to a particular machine or simulation program
 that has no analogue with other machines or simulation programs can be put in as an extension. 
 
 Maintenance of an extension schema falls outside of PALS governance. 
-Extension maintainers can add documentation to this section by submitting a [pull request](#s:contribute).
-Documentation here should be brief and should have a reference to full documentation maintained
+Extension maintainers can add extension documentation submitting a [pull request](#s:contribute).
```
EZoni:
```suggestion
Extension maintainers can add extension documentation by submitting a [pull request](#s:contribute).
```

source/definitions.md
```diff
@@ -80,3 +80,12 @@ of a branch, of constructing the ordered list of lattice elements contained in t
 `Lattice expansion` involves branch expansion along with things like
 calculating the reference energy for all elements.
 
+%---------------------------------------------------------------------------------------------------
+(s:extensions)=
```
EZoni, line 93: Can you double check if this `(s:extensions)` anchor is now dangling?

source/extensions.md
```diff
@@ -6,31 +6,29 @@ For example, information specific to a particular machine or simulation program
 that has no analogue with other machines or simulation programs can be put in as an extension.
```
EZoni, line 6: This is duplicated from definitions.md, which is not ideal, because the two texts will likely end up being out of sync. That might be already the case, given that the text in definitions.md reads
```
An Extension is data added to a PALS lattice that is not part of the PALS standard.
For example, information specific to a particular machine that has no analogue with other
machines can be put in as an extension. A list of "registered" extensions is in the 
[Extension](#c:extension) section.
```
hence uses `PALS lattice` instead of `PALS file` and mentions only `machine` instead of `machine and simulation program`. 

I would recommend having a one-liner in definitions.md and leaving the verbose description only in extensions.md.

source/extensions.md
```diff
@@ -39,13 +37,14 @@ synch_connect:                         # Extension name
...
-Alternatively, extension metadata names may be "registered" using an `extension_labels` attribute
-to be used as the start of the extension.
+Alternatively, extension metadata names may be "registered" using an `extension_labels` attribute.
+Registration means that a PALS parser can validate the extension name.
+Extension metadata names appear at the start (top level) of an extension block.
 The `extension_labels` attribute must appear as a child of the `PALS` root node.
-The `extension_labels` attribute can also be used to register list values.
-`extension_labels` has three components, at least one of them is required:
+The `extension_labels` block can also be used to define enum values.
+The `extension_labels` block has three components, at least one of them is required:
```
EZoni: Here we use both "`extension_labels` block" and "`extension_labels` attribute". Could we use only one wording (either "block" or "attribute") consistently instead?

source/extensions.md
```diff
@@ -39,13 +37,14 @@ synch_connect:                         # Extension name
...
 
-Alternatively, extension metadata names may be "registered" using an `extension_labels` attribute
-to be used as the start of the extension.
+Alternatively, extension metadata names may be "registered" using an `extension_labels` attribute.
+Registration means that a PALS parser can validate the extension name.
+Extension metadata names appear at the start (top level) of an extension block.
 The `extension_labels` attribute must appear as a child of the `PALS` root node.
-The `extension_labels` attribute can also be used to register list values.
-`extension_labels` has three components, at least one of them is required:
+The `extension_labels` block can also be used to define enum values.
```
EZoni: We don't explain how an "enum" is defined versus an ordinary name. From the example, an enum value is just a `names` entry that happens to be matched as a value rather than a key (is it?). I think a sentence making that explicit could help, since  we now state twice that "enums can be defined" but never really show it as a distinct mechanism.

source/definitions.md
DavidSagan: I'm not sure what you mean by "dangling".

source/extensions.md
DavidSagan: I changed `PALS lattice` to `PALS file` and added `machine or simulation program`. There are only two sentences in definitions.md and I don't mind if they get out of synch as long as they are correct.

source/extensions.md
DavidSagan: Changing to `block`.

source/definitions.md
DavidSagan: Replaced `registered` with `documented`.

source/extensions.md
DavidSagan: I added a definition for `enum` in definitions.md and added a link to the def in extensions.md.

DavidSagan:
@EZoni Comments addressed. Please have a look...

source/definitions.md
EZoni: I meant that I'm confused with these `s:` and `c:` anchors and I'm not sure when we use which. Can you remind me? In this specific case, with "dangling" I meant if the `(s:extensions)` anchor is actually referenced/used or if it was meant to be `(c:extensions)`.

DavidSagan: The convention I have been using is `s:` indicates a section (that is, used with a `##` construction) and `c:` indicates a chapter (that is, used with a `#` construction).

source/definitions.md
```diff
@@ -80,12 +80,20 @@ of a branch, of constructing the ordered list of lattice elements contained in t
 `Lattice expansion` involves branch expansion along with things like
 calculating the reference energy for all elements.
 
+%---------------------------------------------------------------------------------------------------
+(s:enum)=
+## Enum
+
+An `enum` key is a parameter that has set of named valid values. For example, the `cavity_type` parameter of
+the [`RFP`](#s:rf.params) parameter group has valid values `STANDING_WAVE` and `TRAVELING_WAVE`.
+The advantage of designating a parameter as an enum means that validation can be done of the parameter value.
```
EZoni, line 90: Regarding

> The advantage of designating a parameter as an enum means that validation can be done of the parameter value.

how's this different from a regular parameter, e.g., say, a simple integer? Validation can be done on the parameter value for a simple integer too, e.g., there can be a constraint about the sign of the integer encoded in the definition of the parameter on the implementation end.

source/definitions.md
```diff
@@ -80,3 +80,20 @@ of a branch, of constructing the ordered list of lattice elements contained in t
...
+
+%---------------------------------------------------------------------------------------------------
+(s:extensions)=
+## Extensions
+
+An Extension is data added to a PALS file that is not part of the PALS standard.
+For example, information specific to a particular machine or simulation program that has no analogue with other
+machines can be put in as an extension. A list of "documented" extensions is in the
```
EZoni: Do we still need the quotes here? I think the word documented is clear enough in itself:

```suggestion
machines can be put in as an extension. A list of documented extensions is in the 
```

source/definitions.md
DavidSagan: Here you are adding possible values. With an integer, you are restricting the range and this would need a different syntax. Similarly, restricting string values would need a different syntax. This could be the subject of a future PR but is not germane to this one.

DavidSagan:
@EZoni Ready for rereview...

source/extensions.md
```diff
@@ -65,23 +64,23 @@ Example:
...
     ...
-    - R23:
-        kind: Rotator   # enum value is an extension name.
+    - R23:              # Another lattice element
+        kind: Rotator   # enum value is a documented extension.
 ```
-The advantage of using `extension_labels` is that single attribute key extensions are possible
-(there does not have to be an `extension` key).
+The advantage of using `extension_labels` is that enums can be defined and 
+single attribute key extensions are possible (there does not have to be an `extension` key).
```
ax3l, line 84: This might be a bit too hard to read and understand.
I would suggest

```suggestion
The advantage of using `extension_labels` is that is can be used to mask attributes at any level.
```

Also the example above does not show an enum...

source/definitions.md
ax3l: I see why you add this: often we have attributes that have a fixed amount of pre-defined values that are allowed. They are usually lists of strings (but could be list-of-any-type). We already use `[enum]` in the standard w/o defining it, e.g.,
https://pals-project.readthedocs.io/en/latest/lattice-element-parameter-groups.html#rfp-rf-parameters

This is a sensible short-hand, we should say at least once that `enum` is an abbreviation for "enumerated type", for the uninitiated.

source/definitions.md
```diff
@@ -80,3 +80,20 @@ of a branch, of constructing the ordered list of lattice elements contained in t
 `Lattice expansion` involves branch expansion along with things like
 calculating the reference energy for all elements.
 
+%---------------------------------------------------------------------------------------------------
+(s:enum)=
+## Enum
+
+An `enum` key is a parameter that has set of named valid values. For example, the `cavity_type` parameter of
```
ax3l:
```suggestion
An enumerated type (`enum`) is a parameter that has set of named valid values.
Typically, this is a list of allowed strings.
For example, the `cavity_type` parameter of
```

source/definitions.md
DavidSagan: Enum is defined in definitions.md and there is link to the def inextensions.md.

source/extensions.md
DavidSagan: The phrasing  `mask attributes` is cryptic to me. Attributes are not being masked (hidden).

source/definitions.md
DavidSagan: Essentially enum values are strings so "Typically, this is a list of allowed strings." should be left out.  See the revised text.

DavidSagan:
@ax3l @EZoni @cemitch99 Ready for rereview...

DavidSagan:
@ax3l Do you have any more comments?

source/extensions.md
qianglbl: @DavidSagan I agree. "mask attributes" is not clear to me.

qianglbl (dismissed):

source/definitions.md
```diff
@@ -78,6 +78,22 @@ An important concept is [`lattice expansion`](#s:lattice.expand) and `branch exp
 Branch expansion is the process, starting from the `root` `BeamLine`
 of a branch, of constructing the ordered list of lattice elements contained in that branch.
 `Lattice expansion` involves branch expansion along with things like
-calculating the reference energy for all elements. Lattice expansion by default occurs
-at the end when the PALS file has been read or when there is a [`expand_lattice`](#s:expand.lat)
-command.
+calculating the reference energy for all elements.
```
EZoni, line 81: Double checking because I might have missed track of the latest updates, was this whole sentence
```
Lattice expansion by default occurs at the end when the PALS file has been read or when there is a [`expand_lattice`](#s:expand.lat) command.
```
removed intentionally as part of this PR or was this an accident of merging/rebasing? I can't seem to find the commit where this was removed.

source/extensions.md
```diff
@@ -39,13 +38,14 @@ synch_connect:                         # Extension name
...
-PALS validation process.
+PALS validation process. Either a dict or a list can be used for the children of `<some_name>`.
 
-Alternatively, extension metadata names may be "registered" using an `extension_labels` attribute
-to be used as the start of the extension.
-The `extension_labels` attribute must appear as a child of the `PALS` root node.
-The `extension_labels` attribute can also be used to register list values.
-`extension_labels` has three components, at least one of them is required:
+Alternatively, extension metadata names may be "documented" using an `extension_labels` block.
+Registration means that a PALS parser can validate the extension name.
```
EZoni: I think now it may be unclear what "Registration" refers to here.

source/extensions.md
```diff
@@ -39,13 +38,14 @@ synch_connect:                         # Extension name
 ```
 In this example, the `synch_connect` dictionary key is the start of the extension. The type of extension is
 `Cornell_CESR_Connect`. Any further entries in `synch_connect` will be excluded from any
-PALS validation process.
+PALS validation process. Either a dict or a list can be used for the children of `<some_name>`.
```
EZoni: I may be missing something or misreading, but this may contradict the syntax template above,
```yaml
<some_name>:
  extension: <extension-type>
  ... extension info ...
```
which shows `extension: <extension-type>` as a child of `<some_name>`. I think one can't have a mapping key (`extension: <extension-type>`) and a sequence (`dict` or `list`) at the same node (the child node). Do we mean that the list form is allowed for the content under the `extension` key?

source/definitions.md
```diff
@@ -78,6 +78,22 @@ An important concept is [`lattice expansion`](#s:lattice.expand) and `branch exp
...
+
+%---------------------------------------------------------------------------------------------------
+(s:enum)=
+## Enum
+
+An `enum` (often called an "enumerated type") key is a parameter that has set of named valid values.
```
EZoni:
```suggestion
An `enum` (often called an "enumerated type") key is a parameter that has a set of named valid values. 
```

source/definitions.md
```diff
@@ -78,6 +78,22 @@ An important concept is [`lattice expansion`](#s:lattice.expand) and `branch exp
...
+calculating the reference energy for all elements.
+
+%---------------------------------------------------------------------------------------------------
+(s:enum)=
+## Enum
+
+An `enum` (often called an "enumerated type") key is a parameter that has set of named valid values. 
+For example, the `cavity_type` parameter of
+the [`RFP`](#s:rf.params) parameter group has valid values `STANDING_WAVE` and `TRAVELING_WAVE`.
+The advantage of designating a parameter as an enum means that validation can be done of the parameter value.
```
EZoni:
```suggestion
The advantage of designating a parameter as an enum is that the parameter value can be validated.
```

source/extensions.md
```diff
@@ -6,31 +6,30 @@ For example, information specific to a particular machine or simulation program
 that has no analogue with other machines or simulation programs can be put in as an extension. 
 
 Maintenance of an extension schema falls outside of PALS governance. 
-Extension maintainers can add documentation to this section by submitting a [pull request](#s:contribute).
-Documentation here should be brief and should have a reference to full documentation maintained
-elsewhere. The advantage of this setup is that since the full documentation is external to the PALS repo,
+Extension maintainers can add extension documentation by submitting a [pull request](#s:contribute).
+If the amount of documentation is "large", it is encouraged that the documentation put in the PALS
+repository should be relatively brief and there should be a reference to full documentation maintained
+elsewhere, if such documentation exists.
```
EZoni: Not a native speaker so I may be wrong, but I thought one usually writes either "it is encouraged that X be brief" or "X should be brief", rather than a mix of the two:
 
```suggestion
If the amount of documentation is "large", the documentation put in the PALS
repository should be relatively brief and there should be a reference to full documentation maintained
elsewhere, if such documentation exists.
```

source/definitions.md
DavidSagan: Intentionally removed since `expand_lattice` is being removed in #293

source/extensions.md
DavidSagan: I have reworded this and removed the word "Registration".

source/extensions.md
DavidSagan: Yes. I have changed this to just be a dict.

source/definitions.md
DavidSagan: Done.

source/definitions.md
DavidSagan: Done.

DavidSagan:
@EZoni Ready for rereview.

source/extensions.md
```diff
@@ -39,13 +39,14 @@ synch_connect:                         # Extension name
...
-PALS validation process.
+PALS validation process. 
 
-Alternatively, extension metadata names may be "registered" using an `extension_labels` attribute
-to be used as the start of the extension.
-The `extension_labels` attribute must appear as a child of the `PALS` root node.
-The `extension_labels` attribute can also be used to register list values.
-`extension_labels` has three components, at least one of them is required:
+Alternatively, extension metadata names may be declared using an `extension_labels` block.
+By declaring  metadata names, a PALS parser is prevented from incorrectly identifying such data as invalid.
```
EZoni:
```suggestion
By declaring metadata names, a PALS parser is prevented from incorrectly identifying such data as invalid.
```

source/extensions.md
```diff
@@ -6,31 +6,31 @@ For example, information specific to a particular machine or simulation program
...
+An extension block can be introduced at any level in a PALS file.
+The syntax is:
 ```{code} yaml
 <some_name>:
   extension: <extension-type>
   ... extension info ...
 ```
-where `<some_name>` is any name (since a given extension type might appear in multiple places in
-the PALS file), and `<extension-type>` is the type of the extension. Example:
+where `<some_name>` is any name containing a dict,
```
EZoni: I think this is slightly more precise (the name doesn't contain the dict per se):
```suggestion
where `<some_name>` is any name whose value is a dict,
```

source/extensions.md
```diff
@@ -39,13 +39,14 @@ synch_connect:                         # Extension name
 ```
 In this example, the `synch_connect` dictionary key is the start of the extension. The type of extension is
 `Cornell_CESR_Connect`. Any further entries in `synch_connect` will be excluded from any
-PALS validation process.
+PALS validation process.
```
EZoni:
```suggestion
PALS validation process.
```

source/definitions.md
```diff
@@ -78,6 +78,22 @@ An important concept is [`lattice expansion`](#s:lattice.expand) and `branch exp
...
-calculating the reference energy for all elements. Lattice expansion by default occurs
-at the end when the PALS file has been read or when there is a [`expand_lattice`](#s:expand.lat)
-command.
+calculating the reference energy for all elements.
+
+%---------------------------------------------------------------------------------------------------
+(s:enum)=
+## Enum
+
+An `enum` (often called an "enumerated type") key is a parameter that has a set of named valid values.
```
EZoni: I think this could be slightly more precise (because it doesn't mix "key" and "parameter"):
```suggestion
An `enum` parameter (often called an "enumerated type") is one whose value must be one of a set of named values.
```

EZoni:
@ax3l @cemitch99 

I think I'm close to reaching my final review round, so this might be a good time for you to review the PR as well.

source/extensions.md
DavidSagan: Whanged to: " where <some_name> is an arbitrary key whose value is a dictionary"

DavidSagan:
@EZoni Ready again and thanks for your proofreading!

source/extensions.md
```diff
@@ -6,31 +6,31 @@ For example, information specific to a particular machine or simulation program
...
+An extension block can be introduced at any level in a PALS file.
+The syntax is:
 ```{code} yaml
 <some_name>:
   extension: <extension-type>
   ... extension info ...
 ```
-where `<some_name>` is any name (since a given extension type might appear in multiple places in
-the PALS file), and `<extension-type>` is the type of the extension. Example:
+where <some_name> is an arbitrary key whose value is a dictionary,
```
EZoni: Backticks were dropped:

```suggestion
where `<some_name>` is an arbitrary key whose value is a dictionary,
```

EZoni (changes requested):

EZoni (approved):
@ax3l @cemitch99 

After a few rounds of review, this PR looks good to me, so I'm approving it. Do you have any comments we should take into account before we merge?

cemitch99 (approved):

qianglbl (approved):

ax3l (commented):
This looks ok to me.

Claude has a few more details but I find them to be minor.
<details>

## Findings

### 1. Premature removal — settle before merge (`source/definitions.md:82`)

The deleted sentence ("Lattice expansion by default occurs at the end when the PALS file has been read
or when there is an `expand_lattice` command") was removed on the grounds that #293 drops
`expand_lattice` — but **#293 is still open**, and `(s:expand.lat)=` is live at
`source/lattice-construction.md:256`. Merging this first leaves `main` with an `expand_lattice`
command section and no statement anywhere of *when* expansion happens.

*Suggested fix:* keep a command-agnostic half — "Lattice expansion by default occurs after the PALS
file has been read." — and let #293 remove the rest.

### 2. Glossary list not updated (`source/definitions.md:11-25`)

The file opens with a bullet glossary (`Branch`, `Lattice expansion`, …). The PR adds `## Enum` and
`## Extensions` sections but no matching `- Enum:` / `- Extension:` bullets, so the two halves of the
file now disagree about what is defined.

### 3. Anchor conventions

Every other section in `definitions.md` uses `s:*.intro` (`s:elements.intro`, `s:expansion.intro`);
the new ones are bare `s:enum` / `s:extensions`. Also `#s:extensions` has no inbound links —
`extensions.md` links to `#s:enum` but never back to the new Extensions definition.

### 4. Duplicated definition will drift (`source/definitions.md:87-94`)

The new Extensions section copies the opening paragraph of `extensions.md` but drops "or simulation
programs", so the two copies already differ on day one. It also says "A list of documented extensions
is in the [Extension](#c:extension) section" while pointing at the whole chapter — the list itself is
at `#s:extension-list`.

### 5. Lost constraint (`source/extensions.md:25`)

"at any dictionary entry within `PALS`" became "at any level in a PALS file".
`source/fundamentals.md:55` states that anything outside the `PALS` node is ignored by parsers, so
"within the `PALS` node" should stay.

### 6. Still-confusing sentence (`source/extensions.md:46`)

"Extension metadata names appear at the start (top level) of an extension block" is circular, and it
fights the closing sentence about single-attribute extensions — in the example `data_base: Q1_czt` is
a scalar, not a block.

*Suggested wording:* "A declared name, prefix, or suffix marks the key where extension data begins;
its value may be a scalar or a dictionary."

### 7. Design question (`source/extensions.md:49,66-72`)

Enum values and key names share one `names:` dict — `Rotator` is disambiguated only by a prose
comment. Since the stated advantage of enums is *validation*, a parser cannot distinguish a typo'd key
from a declared enum value. Was a separate `enum_values:` component considered?

Related: `suffixes` is listed as one of the three components but is never demonstrated in the example.

## Bottom line

A genuine readability improvement over the current text; approve. Only finding #1 is worth blocking
on — it is a real gap on `main` if this lands before #293. Findings #2–#4 are quick follow-ups. #7 is
a substantive question worth answering before the syntax hardens, since it affects schema validation
(the in-progress `schema/` LinkML work does not implement `extension_labels` at all yet, so nothing is
pinned down on the implementation side).
</details>

ax3l (approved):

PR #233, opened 2026-06-29
RBend and SBend to Bend

DavidSagan:
PALS supports more bend geometries than "rectangular" and "sector". So trying to fit bends into "RBend" or SBend" categories does not make much sense. In fact, there is no difference between RBend or SBend other than the name. The bend geometry is selected by the settings of the `ref_coords` component of `BendP` independent of whether RBend or SBend is used. 

Given the above, this PR does away with RBend and SBend and replaces them by Bend. 
Another advantage is that with this change people will not be confused into thinking that PALS "RBend" or SBend" is equivalent to MAD or Bmad or Elegant "RBend" or SBend".


source/lattice-element-kinds.md
```diff
@@ -38,11 +38,9 @@ ack1:
 
 %---------------------------------------------------------------------------------------------------
 (s:bend)=
-###  Bend Elements: RBend and SBend
+###  Bend Element
 
-Dipole bend. There are two kinds of bends depending upon the "logical shape". 
-The `RBend` element has a "rectangular" logical shape and the `SBend` element has a "sector"
-logical shape.
+Dipole bend.
```
cemitch99: A minor suggestion:  Most of the element kinds include a brief one-sentence description.  Should we include a description of what is meant by "dipole bend"?   For example:  the dominant contribution to the magnetic field is a constant magnetic field transverse to the direction of motion, responsible for bending the reference orbit (or something similar).

DavidSagan: Added description.

DavidSagan:
@cemitch99 Ready for rereview...

source/parameters/bend.md
```diff
@@ -9,31 +9,34 @@ length `length` parameter.
...
+                           # [enum] Sets how multipoles are calculated.
+  ref_geometry: arc        # [enum] Reference bend geometry.
+  rho_ref: null            # [m]s Reference bend radius.
   tilt_ref: 0              # [radian] Reference tilt
 ```
 
-The geometry of a bend in [body](#s:coords).  coordinates is shown in {numref}`f:bend`. 
-In body coordinates the rotation axis is, by definition, always the {math}`y`-axis.
+The geometry of a bend in [body](#s:coords) coordinates is shown in {numref}`f:bend`. 
+In body coordinates, the rotation axis is, by definition, always the {math}`y`-axis.\footnote
```
cemitch99: This footnote does not appear to render correctly.

source/parameters/bend.md
```diff
@@ -50,15 +53,15 @@ The red curve is the reference curve.
...
 `g_ref`, `angle_ref`, `rho_ref`, `e1`, `e2`, `e1_rect`, and `e2_rect` are all positive.
 
 B) Same as (A) but with a negative bend angle. For the geometry shown,
 `g_ref`, `angle_ref`, `rho_ref`, `e1`, `e2`, `e1_rect`, and `e2_rect` are all negative.
 
-C) Bend geometry for `ref_coords` set to `entrance_coords`.
+C) Bend geometry for `ref_geometry` set to `entrance_coords`.
 
-D) Bend geometry for `ref_coords` set to `exit_coords`.
+D) Bend geometry for `ref_geometry` set to `exit_coords`.
```
cemitch99, line 56: In the sequence of subfigure captions (A-D), label D is not indented at the same level as the others in the rendered document.

DavidSagan: That is because the captions are centered.

source/parameters/bend.md
DavidSagan: Removed footnote and moved text to figure caption.

cemitch99 (approved):

PR #237, opened 2026-07-06
Controller

### LINKED ISSUE #235: Define controller elements
Controller elements (called `overlays` and `groups`) in Bmad are very useful so it would be good to have the analogous construct in PALS. A PR will be created... Note: The Bmad manual has documentation on controller elements.

DavidSagan:
With Bmad, controllers are widely used in simulations of a wide range of machines at a number of laboratories This is due to the fact that what controllers simulate are basic building blocks for virtually all accelerators. While deferred expressions can be used in place of controllers, controllers add a level of organization and modularity that make them useful constructs.  The ability to add documentation to a controller and the ease of being able to search and sort by controller name are features that deferred expressions lack.

Close #235

### DIFF
--- a/source/miscellaneous.md
+++ b/source/miscellaneous.md
@@ -58,12 +58,12 @@ facility:
         cur1: 0.023
         cur2: cur1 / c_light
         ...
-      controls:                        
+      controls: 
         - parameter: Qa.*>MagneticMultipoleP.Ks2L  # Parameter control specification.
           expression: 0.075*sin(cur1) + 0.3*cur2
         - parameter: ...                           # Another specification
           expression: ...           
-       ...
+        ...
 ```
 In this example, a controller called `ps27` has two variables used to control parameters called
 `cur1` and `cur2`. Initial values can be set for variables. The default value of a variable is zero.
@@ -71,12 +71,9 @@ The parameters controlled are set by the `controls` component. Each entry has a
 `parameter` component giving the [regular expression](#s:parameter.matching) to 
 match parameters to and an `expression` component that gives the arithmetic expression
 used in setting the parameters matched to. 
-In the above example, the first `controls` group will match to the `Ks2L` 
-component of all elements whose name begins with `Qa`. 
-
-
- In this example, that expression is
-`0.075*sin(cur1) + 0.3*cur2`.
+In the above example, the first `controls` entry will match to the `Ks2L` 
+component of all elements whose name begins with `Qa`, and the associated expression
+is `0.075*sin(cur1) + 0.3*cur2`.
 
 Controllers can control the variables of other controllers. The syntax for a controller 
 variable is:
@@ -138,7 +135,7 @@ facility:
       control_type: ABSOLUTE
       variables:
         cur: 0.023
-      controls: 
+      controls:
         - parameter: a_kicker>MagneticMultipoleP.Kn0
           expression: 0.075*sin(cur)
         ...
@@ -148,10 +145,10 @@ facility:
       control_type: ABSOLUTE
       variables:
         cur: 0.044
-      controls: 
-          - parameter: a_kicker>MagneticMultipoleP.Kn0
-            expression: 0.123*cur
-          ...           
+      controls:
+        - parameter: a_kicker>MagneticMultipoleP.Kn0
+          expression: 0.123*cur
+        ...           
 ```
 Here two controllers control the `Kn0` parameter of the `a_kicker` element.
 In a case where there are multiple `ABSOLUTE` controllers controlling the same parameter,
--- a/source/miscellaneous.md
+++ b/source/miscellaneous.md
@@ -60,7 +60,7 @@ facility:
         description: Model Mitsubishi 800KL
       variables:
         cur1: 0.023
-        cur2: cur1 / c_light
+        cur2: 1e8 / c_light
         ...
       controls:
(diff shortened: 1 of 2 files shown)


mattsignorelli:
Just to continue the discussion in #235 here:
Deferred expressions are meant to embody physical dependencies such as power supplies. And specifying summation with them is no problem. My only concern is that having several different ways of specifying the same thing in the standard may be suboptimal. All that `Controller` achieves can currently be achieved using deferred expresions

DavidSagan:
@mattsignorelli  My experience is that in practice controllers can simplify machine control organization.  For example, how would you handle the relative controller chromaticity example with deferred expressions? (actually I'm not sure deferred expressions can really handle this case.)

mattsignorelli:
> @mattsignorelli My experience is that in practice controllers can simplify machine control organization. For example, how would you handle the relative controller chromaticity example with deferred expressions? (actually I'm not sure deferred expressions can really handle this case.)

In SciBmad this would look like:

```julia
k2 = 0.33
command = 0
dk2 =  DefExpr(()->5.62*command + 0.02*command^2 )

S1 = Sextupole(Kn2=k2 + dk2)
```

I attached a file including the sextupole family knobs (which automatically keep chromaticity to +1 in both planes) that I am using now in the ESR.

[Archive.zip](https://github.com/user-attachments/files/29714604/Archive.zip)

DavidSagan:
> > @mattsignorelli My experience is that in practice controllers can simplify machine control organization. For example, how would you handle the relative controller chromaticity example with deferred expressions? (actually I'm not sure deferred expressions can really handle this case.)
> 
> In SciBmad this would look like:
> 
> ```julia
> k2 = 0.33
> command = 0
> dk2 =  DefExpr(()->5.62*command + 0.02*command^2 )
> 
> S1 = Sextupole(Kn2=k2 + dk2)
> ```
> 
> I attached a file including the sextupole family knobs (which automatically keep chromaticity to +1 in both planes) that I am using now in the ESR.
> 
> [Archive.zip](https://github.com/user-attachments/files/29714604/Archive.zip)

First what you show acts differently from what the relative controller does. And how would you implement this in a PALS file (remember, PALS just defines expressions but applying them after the lattice is read in is up to the program. So the equivalent of `S1 = Sextupole(Kn2=k2 + dk2)` would not be in the PALS file).

mattsignorelli:
> First what you show acts differently from what the relative controller does. 

Could you clarify the differences specifically? I must be misunderstanding your text in this PR.

> And how would you implement this in a PALS file (remember, PALS just defines expressions but applying them after the lattice is read in is up to the program. So the equivalent of S1 = Sextupole(Kn2=k2 + dk2) would not be in the PALS file).

So PALS does not have deferred expressions?

DavidSagan:
PALS does have deferred expressions. PALS defines what you could call the "initial lattice". Variations of parameters happens after the PALS file is read and this variation is not part of the PALS file. Controllers made parameter variation straight forward in a way that bare expressions do not. In your above example, the delta `dk2` should be computed at the difference between the beginning value of command and the ending value. Also you only show control of one sextupole. How do you write this when there are N sextupoles to control?

mattsignorelli:
> PALS does have deferred expressions. PALS defines what you could call the "initial lattice". Variations of parameters happens after the PALS file is read and this variation is not part of the PALS file. 

I think we might have disagreeing definitions of "deferred expressions".  What I am referring to the deferred expression as defined in MAD-X (using `:=`) , MAD-NG, Xsuite, and SciBmad. The whole point is that you can change parameters and have their interdependencies hardwired in. If PALS does not support variation , then PALS does not have deferred expressions

DavidSagan:
Currently, PALS is setup to define an initial lattice. Variation of parameters is left to the simulation program. There is no concept in PALS as to immediate evaluation vs. deferred evaluation. Essentially the program knows about all the equations in the PALS file and is free to vary any parameter that the User wants to vary.

DavidSagan:
Note: Reference to the "delayed expression" definition will not be valid until #240 PR is merged.

cemitch99 (dismissed):

source/miscellaneous.md
```diff
@@ -40,3 +40,134 @@ For sets that only have a value, an alternative compact form has the syntax:
...
+```{code} yaml
+facility:
+  - ps27:                             # Controller name
+      kind: Controller
+      control_type: ABSOLUTE           # Or RELATIVE.
+      MetaP:
+        description: Model Mitsubishi 800KL
+      variables:
+        cur1: 0.023
+        cur2: cur1 / c_light
```
ax3l: @DavidSagan I am just thinking through an implementation here:
Would variables here be allowed to reference or inherit another controller or element variable in it?

If we allow this (which seems natural in the existing PALS constructs we allow), we should define the precedence of evaluation (in https://github.com/pals-project/pals/pull/240#discussion_r3574489341)

DavidSagan: I'm not quite sure what you man by "Would variables here be allowed to reference or inherit another controller or element variable in it?" A controller can control the variables of another controller so there can be a hierarchy  of controllers (loops are forbidden). The hierarchy can be determined by the parser and passed to the calling program. Evaluation is from the top down. I will add this info to the PR...

DavidSagan: I have add documentation on this. Note that, to keep things sane, controller expressions cannot contain lattice parameters.

cemitch99 (approved):

PR #238, opened 2026-07-06
Added `location` and `history` to `MetaP`.

DavidSagan:
Added `location` and `history` to `MetaP`.

### DIFF
--- a/source/parameters/meta.md
+++ b/source/parameters/meta.md
@@ -9,8 +9,8 @@ MetaP:
   ID: ""            # [string] An identification string.
   label: ""         # [string] A label string
   description: ""   # [string] A descriptive string
-  location: ""      # [string] To describe the location of an object.
-  history: ""       # [string] Object's history.
+  location: ""      # [string] To describe the location of element.
+  history: ""       # [string] Element's history. 
 ```
 In addition to an element's `name`, these four strings can be used for pattern matching
 when trying to locate all elements of a given type.
@@ -28,6 +28,10 @@ quad1:                  # user-defined name
   kind: Quadrupole      # [string] element switch
   MetaP:
     ID: 0137-85
+    location: East transfer line
+    history:
+      - 2022-04-01: Fixed water leak to vacuum.
+      - 2023-06-24: Fixed short in main bus bar.
     manufacture:        # Custom information
       - date: 2017-03-07
       - installed: 2018-01-23
--- a/source/parameters/meta.md
+++ b/source/parameters/meta.md
@@ -21,7 +21,7 @@ Example metadata could be blueprint information, information on power supply con
 
 Components of `MetaP` are not limited to being simple strings or numbers but can be complex 
 structures. The information contained in `MetaP` should be restricted to information that 
-does not affect simulations. Also program specific (as opposed to machine specific) information
+does not affect simulations. Also, program specific (as opposed to machine specific) information
 should be stored elsewhere.
 ```{code} yaml
 quad1:                  # user-defined name
--- a/source/parameters/meta.md
+++ b/source/parameters/meta.md
@@ -2,33 +2,33 @@
 (s:meta.params)=
 ## MetaP: Metadata Parameters
 
-`MetaP` has four standard components
+`MetaP` has six standard components:
 ```{code} yaml
 MetaP:
   alias: ""         # [string] An alternate name for the element.
   ID: ""            # [string] An identification string.
   label: ""         # [string] A label string
   description: ""   # [string] A descriptive string
+  location: ""      # [string] To describe the location of an object.
+  history: ""       # [string] Object's history.
 ```
 In addition to an element's `name`, these four strings can be used for pattern matching
 when trying to locate all elements of a given type.
 
-Besides the four standard strings, the `MetaP` parameter group can be used to store metadata 
+Besides the six standard strings, the `MetaP` parameter group can be used to store metadata 
 that describes the lattice element but is not part of the PALS standard. 
-Such data is necessarily program dependent.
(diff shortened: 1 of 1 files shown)


source/parameters/meta.md
```diff
@@ -2,33 +2,34 @@
...
 
-`MetaP` has four standard components
+`MetaP` has six standard components:
 ```{code} yaml
 MetaP:
   alias: ""         # [string] An alternate name for the element.
   ID: ""            # [string] An identification string.
   label: ""         # [string] A label string
   description: ""   # [string] A descriptive string
+  location: ""      # [string] To describe the location of an object.
```
EZoni: An alternative docstring that follows the noun-phrase style of the other components and includes some examples to define what "location" means here (is this what "location" means here?):

```suggestion
  location: ""      # [string] Physical location of the object (e.g., building, room, or facility position).
```

source/parameters/meta.md
```diff
@@ -2,33 +2,34 @@
...
+  history: ""       # [string] Object's history.
```
EZoni: An alternative docstring that includes some examples to define what "history" means here (is this what "history" means here?):
```suggestion
  history: ""       # [string] Record of notable events in the object's lifecycle (e.g., manufacture, installation, maintenance).
```

source/parameters/meta.md
```diff
@@ -2,33 +2,34 @@
...
+Besides the six standard strings, the `MetaP` parameter group can be used to store metadata 
 that describes the lattice element but is not part of the PALS standard. 
-Such data is necessarily program dependent.
 Example metadata could be blueprint information, information on power supply connections, etc.
 
 Components of `MetaP` are not limited to being simple strings or numbers but can be complex 
 structures. The information contained in `MetaP` should be restricted to information that 
-does not affect simulations. Custom, program specific information should be stored elsewhere.
-Example:
+does not affect simulations. Also program specific (as opposed to machine specific) information
```
EZoni: I think I was erroneously parsing this "also" as "information that is additionally program-specific", but if that's not what this means here, maybe we could add a `,` to make it clearer for non-native speakers:

```suggestion
does not affect simulations. Also, program specific (as opposed to machine specific) information
```

EZoni (changes requested):

source/parameters/meta.md
DavidSagan: See my changes

source/parameters/meta.md
DavidSagan: See my changes.

DavidSagan:
@EZoni Ready for rereview...

source/parameters/meta.md
```diff
@@ -2,33 +2,38 @@
...
 ```{code} yaml
 MetaP:
   alias: ""         # [string] An alternate name for the element.
   ID: ""            # [string] An identification string.
   label: ""         # [string] A label string
   description: ""   # [string] A descriptive string
+  location: ""      # [string] To describe the location of element.
+  history: ""       # [string] Element's history. 
 ```
 In addition to an element's `name`, these four strings can be used for pattern matching
```
EZoni: Can `location` and `history` be used for pattern matching too?

```suggestion
In addition to an element's `name`, these six strings can be used for pattern matching
```

source/parameters/meta.md
```diff
@@ -2,33 +2,38 @@
...
+quad1:                  # user-defined name
   kind: Quadrupole      # [string] element switch
   MetaP:
     ID: 0137-85
-    history:            # Custom information
-      - manufacture: 2017-03-07
+    location: East transfer line
+    history:
+      - 2022-04-01: Fixed water leak to vacuum.
+      - 2023-06-24: Fixed short in main bus bar.
```
EZoni, line 35: In this example `history` is a list, not a string, while the YAML schema block above declares it as a string.

source/parameters/meta.md
```diff
@@ -2,33 +2,38 @@
...
 
-`MetaP` has four standard components
+`MetaP` has six standard components:
 ```{code} yaml
 MetaP:
   alias: ""         # [string] An alternate name for the element.
   ID: ""            # [string] An identification string.
   label: ""         # [string] A label string
   description: ""   # [string] A descriptive string
+  location: ""      # [string] To describe the location of element.
```
EZoni: I would vote for style consistency with the `history` docstring here:

```suggestion
  location: ""      # [string] Element's location.
```

source/parameters/meta.md
```diff
@@ -2,33 +2,38 @@
...
 Components of `MetaP` are not limited to being simple strings or numbers but can be complex 
 structures. The information contained in `MetaP` should be restricted to information that 
-does not affect simulations. Custom, program specific information should be stored elsewhere.
-Example:
+does not affect simulations. Also, program specific (as opposed to machine specific) information
+should be stored elsewhere.
 ```{code} yaml
-quad1:                  # [string] user-defined name
+quad1:                  # user-defined name
   kind: Quadrupole      # [string] element switch
```
EZoni: If `[string]` was removed intentionally from the docstring of `quad1` in this example, then maybe we should remove it from the docstring of `kind: Quadrupole` too?

```suggestion
  kind: Quadrupole      # element switch
```

Not sure what convention we follow in general (type reminder in examples or not).

EZoni (changes requested):

source/parameters/meta.md
DavidSagan: Good catch! `history` (and two other components) should not have been marked as a string. I have revised the documentation.

DavidSagan:
@EZoni Ready for rereview...

EZoni (approved):

cemitch99 (approved):

PR #240, opened 2026-07-08
`variables` and expressions: Introduce

### LINKED ISSUE #239: Constants vs Variables
Right now a PALS file defines an initial set of parameters and values (which can be expressions) and after the PALS file is parsed, the reading program can do anything it wants in terms of varying parameters. That is, PALS does not make any distinction between parameters whose value is an expression and should be held constant ("immediate expression evaluation") and parameters that can be varied ("deferred expressions"). 

The question is should the PALS standard be modified so that expressions can designated as deferred or immediate evaluation? If so, what would be the syntax for this? Also currently PALS has constants that can be defined (but not reset within the PALS file). Should there be variables as well that could be set to different values within a PALS file?

DavidSagan:
- Document `variables` and expressions.
- Introduced delayed expressions

Closes #239

Note:  The controller documentation sits on another PR that has not yet been merged so the reference to the controller section is not yet valid.


source/fundamentals.md
```diff
@@ -196,15 +196,25 @@ n_avogadro                # [-] Avogadro's constant
 The `classical_radius_factor` is a useful number when converting a formula that involves the classical
 electron or proton radius to a formula for something other than an electron or proton.
 
-Other constants may be defined using `constant` as the `kind`.
-The parameters associated with a constant are:
+Other constants and variables may be defined using `constant` or `variable` as the `kind`.
+The difference between a constant and a variable is that a constant is not supposed to be
+varied after a program reads a PALS file while a variable can vary. Since what happens
```
jlvay: I find lines 201-206 very confusing. Also, something like:
- my_var
    kind: constant
    value: expr(...)

should be illegal, hence PALS should not treat constant and variables the same.

mattsignorelli:
Only marking an expression with `expr` makes "nesting" of deferred expressions undefined. E.g. what value would `c` have in this example if `a` is changed

```
a: 0
b: expr(a)
c: b + 1
```

DavidSagan:
@mattsignorelli c would have the value 1. Nothing undefined here. Think of `expr()` as equivalent to `:=` in MAD and not using `expr()` as equivalent to `=`.

source/fundamentals.md
DavidSagan: Documentation has been modified. Please take a look.

source/fundamentals.md
```diff
@@ -213,27 +216,46 @@ PALS:
...
-      - const_c:: pi * const_a  # Can use expressions.
+      - const_c: pi * const_a   # Can use expressions.
+      ...
+  - variables:
+      - var_a: var_val_a        # Define var_a
       ...
 ```
+The difference between a constant and a variable is one of intent of the PALS file creator.
+That is, a constant is not supposed to be
+varied after a program reads a PALS file while a variable can vary.
```
jlvay: Perhaps include examples? E.g.:

That is, a constant (e.g., the strength of a permanent magnet) is not supposed to be varied after a program reads a PALS file while a variable (e.g., the strength of a magnet that varies during operations) can vary.

jlvay (dismissed):

source/fundamentals.md
DavidSagan: OK I put in an example.

source/fundamentals.md
```diff
@@ -263,3 +287,62 @@ mass_of(A)               # Mass of particle A
...
+as listed in the [](#s:functions) section as well as [constants and variables](#s:constants). 
+Standard math symbols `+` plus, `-` minus, `*` times, `/` divide, and `^` power, along with
+parentheses `(...)`, are valid.
+Example:
+```{code} yaml
+- variables:
+    a_var: 3.75e7 / c_light^2
+    b_var: -0.34
+
+cleo:
```
ax3l: missing `-` in the list?

source/fundamentals.md
```diff
@@ -263,3 +287,62 @@ mass_of(A)               # Mass of particle A
 charge_of(A)             # Charge, in units of the elementary charge, of particle A 
 anomalous_moment_of(A)   # Anomalous magnetic moment of particle A
 ```
+The `ran_gauss(sig_cut)` function produces a Gaussian distributed random number with unit sigma
+and zero mean modified so that all random numbers will be in the range `[-sig_cut, sig_cut]`.
+ 
+When `random` or `ran_gauss` functions are used in a single expression that sets multiple parameters,
```
ax3l: If we have `random` I would also use `random_gauss`:
```suggestion
The `random_gauss(sig_cut)` function produces a Gaussian distributed random number with unit sigma
and zero mean modified so that all random numbers will be in the range `[-sig_cut, sig_cut]`.
 
When `random` or `random_gauss` functions are used in a single expression that sets multiple parameters,
```

source/fundamentals.md
```diff
@@ -251,9 +275,9 @@ atanh(x), acoth(x)       # Hyperbolic arc tangent and cotangent
 abs(x)                   # Absolute Value
 factorial(n)             # Factorial
 random()                 # Uniform random number between 0 and 1
-ran_gauss()              # Gaussian distributed random number
-ran_gauss(sig_cut)       # Gaussian distributed random number
-int(x)                   # Nearest integer with magnitude less then x
+ran_gauss()              # Gaussian distributed random number with unit sigma and zero mean
+ran_gauss(sig_cut)       # Same ran_gauss() but with a probability tail cutoff.
```
ax3l: If we have `random` I would also use `random_gauss`:
```suggestion
random_gauss()              # Gaussian distributed random number with unit sigma and zero mean
random_gauss(sig_cut)       # Same random_gauss() but with a probability tail cutoff.
```

source/fundamentals.md
```diff
@@ -196,15 +197,17 @@ n_avogadro                # [-] Avogadro's constant
...
 ```{code} yaml
   absolute_error: 0       # Absolute error.       
   relative_error: 0       # Relative error.
-  value                   # Constant value.
+  value                   # Constant or variable value.
 ```
 If both `absolute_error` and `relative_error` are specified, 
 the true error is `absolute_error + relative_error * |value|`.
+
+Constants and variables must be defined directly under the `PALS` node or the `facility` node.
```
ax3l, line 210: Adding here when they are evaluated would be useful (i.e., to make the parser implementations clear).

Because variables can be re-set later on, just evaluating them upfront will not do the trick, and we need to specify their evaluation order to lattice reading, lattice expansion, and other function and expression evaluations here cleanly.

DavidSagan: See the discussion further down.

source/fundamentals.md
DavidSagan: Fixed.

source/fundamentals.md
DavidSagan: Changed.

source/fundamentals.md
DavidSagan: Changed.

source/fundamentals.md
```diff
@@ -263,3 +287,62 @@ mass_of(A)               # Mass of particle A
...
+Immediate evaluation expressions are expressions that, once a PALS file is read in by a program,
+are meant to be immediately evaluated and never reevaluated. That is, what is stored is the
+value of the expression and the expression used to compute the value can be ignored thereafter. 
+Delayed evaluation on the other hand are meant to be retained by the program and, if any
+variables in the expression change, the program should update the value automatically.
+Delayed evaluation expressions are denoted in a PALS file using the `expr(...)` construct,
+as in the example above, where the expression for `Kn1` (`3.74 * a_var`) is a delayed
+expression. The `expr(...)` construct must enclose the entire expression. That is, something
+like `3.74 * expr(a_var)` is invalid. Note that whether an expression is delayed or immediate is
+determined solely by whether the expression is wrapped in `expr(...)` or not. The fact
```
ax3l, line 341: In order to implement this unambiguous, we need to clarify and document: When are the delayed expressions evaluated?
What is their evaluation/concretization precedence to:
- includes
- lattice expansion
- `set`
- `Controller` evaluations
- inheritance/referencing of other element properties?

ax3l:
The introduction of variables and (immediate) expressions looks ok to me.

We skipped over this in the meeting: but can we please take 15-20min writing down the strong motivating use cases (needs) for globally allowing RHS **delayed expressions**? (More than, "but MAD always had them".) I fully agree with the `Controller` use in #237, which contains the syntax and implications on parsing a PALS file.

Deferred MAD(-X) calls are computationally so-called _lazy evaluations_, similar to Excel spreadsheet `=A1*2`. They are always-evaluated functions referring to global variables. xdeps implements it as a reactive dependency graph, MAD-NG as Lua closures (functions that are called each time).

As far as I am aware, there are two major use cases:
1. Scaling segments/lines. This could also be done with a variable defined upfront instead of at the end of a file.
2. Error/random-value generation. The deferred expression on the RHS is then a random number generator.

Since we introduce `Controller`, we could use them like clearly parametrized _functions_ (with their `variables`) and keep the deferred lookup depth contained.

## Where my concern comes from

1. SWE concern: It could become very complicated to unambiguously implement a parser if we allow too much dynamic back and forth. We already have inheritance, references, variables, `set`, expressions, `Controller` at our disposal defined in PALS.
  There is a risk making PALS an interpreter instead of a data standard with too many dynamic features, which complicates implementations/portability/reproducibility.
3. Hand-waving concern: Users often cite deferred variables as a source of bugs and confusion.

<details>
<summary>Potential outcome</summary>

My guess is that if we spend time listing the use cases, we will see that either reordering a file with immediate expressions and variables, or parametrizing a PALS lattice file (e.g., Jinja templating a lattice file), or parametrizing a generator script in a scripting language will be:
- more clear
- more well known
- less confusing
- play better with the rest of the scientific ecosystem
</details>

source/fundamentals.md
```diff
@@ -263,3 +287,62 @@ mass_of(A)               # Mass of particle A
...
+that an expression may contain a variable (see the `length` expression in the above example), 
+is not a consideration.
+The exception is that [controller](#s:controller) expressions are always considered to be delayed. 
+Also note that what a program actually does is outside of the PALS purview. The designation of
+immediate or delayed is only a "hint" to the program of what the lattice designer intends.
```
ax3l, line 346: Making something as essential as global `expr()`-support in a reader optional could hurt PALS file interoperability?
Interoperability is the primary purpose of PALS, the same file would produce different physics if `expr` are ignored.

source/fundamentals.md
```diff
@@ -251,9 +275,9 @@ atanh(x), acoth(x)       # Hyperbolic arc tangent and cotangent
 abs(x)                   # Absolute Value
 factorial(n)             # Factorial
 random()                 # Uniform random number between 0 and 1
-ran_gauss()              # Gaussian distributed random number
-ran_gauss(sig_cut)       # Gaussian distributed random number
-int(x)                   # Nearest integer with magnitude less then x
+random_gauss()           # Gaussian distributed random number with unit sigma and zero mean
+random_gauss(sig_cut)    # Same random_gauss() but with a probability tail cutoff.
```
ax3l, line 279: We need to introduce seed-control for all existing and new RNGs or we will have issues with reproducibility and interoperability.

DavidSagan:
Delayed expressions are needed to simulate things like the control system. This is needed with the lattice information, for example, for digital twins. Trying to separate control system info from the lattice info actually makes things more complicated. Note that in terms of implementation, MAD uses "bare" delayed expressions but does not have a controller construct . On the other hand, Bmad has controllers (which can be thought of as "packaged" delayed expressions) but not bare delayed expressions. I've never felt the need for bare delayed expressions in Bmad (although I suspect Scott might have a different opinion on this).

Note: How a program does evaluation in terms of lazy or immediate is not relevant to PALS. PALS essentially is defining an initial state.

source/fundamentals.md
DavidSagan: Interoperability should not be an issue. Everything can be computed by the PALS parser and given to the calling program so the calling program can just use the values and ignore any expressions. Expressions are just for programs that can use them and want to vary the variables using the expressions. Note: Any program can do optimization on any set of parameters for stuff like lattice design independent of any expressions.

source/fundamentals.md
DavidSagan: I will leave that to a future PR...

source/fundamentals.md
DavidSagan: I have added a new section on evaluation order.

DavidSagan:
> 2\. Hand-waving concern: Users often cite deferred variables as a source of bugs and confusion.

Yes I have heard anecdotally that bare deferred variables can be a headache. This is one reason why I like controllers over bare deferred variables in that the bare deferred variable names live in the global namespace while controller variables are in the namespace of the controller name and it is not allowed to have multiple controllers of the same name. It is therefore not possible to redefine a controller variable by accident.

DavidSagan:
> As far as I am aware, there are two major use cases:
> 
> 1. Scaling segments/lines. This could also be done with a variable defined upfront instead of at the end of a file.
> 2. Error/random-value generation. The deferred expression on the RHS is then a random number generator.

A typical Cornell CESR lattice file has nearly 300 controllers for defining bumps, chromaticity knobs, etc. And CESR is small being only 800 m in circumference. None of these controllers involve your two categories. There are many use cases for expressions.

mattsignorelli:
While I agree that lazy vs. eager evaluation is irrelevant to PALS, I still think this decision should be considered carefully. I also think it would be good to involve @ldeniau in this conversation. Also will ping @jsberg-bnl 

SciBmad also implements deferred expressions as closures. My experience both personally and anecdotally from at least some colleagues is the opposite of yours - we find deferred expressions to much easier to reason about than applying top-down "Controllers", as in the classical Bmad sense.

Also the variables do not need to live in any global scope. In SciBmad I have "contexts" where you can define control variables and can swap and evaluate them as needed. The syntax looks like this, in Python:

```python
c = Context(a = 0.36)
qd = Quadrupole(Kn1L = DefExpr(lambda c : -c.a))
```

Then each beamline stores a context which can be mutated/swapped out
```python
bl = Beamline([qd], context=c)
bl.line[0].Kn1L # returns 0.36
c.a = 0.37
bl.line[0].Kn1L # returns 0.37
```

mattsignorelli:
> We skipped over this in the meeting: but can we please take 15-20min writing down the strong motivating use cases (needs) for globally allowing RHS delayed expressions? (More than, "but MAD always had them".) 

Taking a sentence from Laurent's MAD-NG paper (https://arxiv.org/pdf/2412.16006):

> It retains the full set of deferred expressions – about a hundred thousand for the LHC (Large Hadron Collider, 27 Km) and HL-LHC (High Luminosity LHC upgrade) lattices – to define functional knobs, circuits, and core lattice logic, preserving the design flexibility and functionality typical of the MAD series codes

DavidSagan:
> Also the variables do not need to live in any global scope.

This is very true but note that as of now there are no namespaces in PALS.

source/fundamentals.md
ax3l: Awesome, I'll add questions inline for clarity.

source/fundamentals.md
```diff
@@ -263,3 +287,84 @@ mass_of(A)               # Mass of particle A
...
+- A "merged" tree is constructed where any include files are merged into the root file tree.
+
+- The controllers are evaluated last as a group. Controllers with a `RELATIVE` `control_type` do
+not initially affect any parameters (it is only if the program reading the PALS file later varies a 
+RELATIVE controller variable that there is any effect). The ABSOLUTE controllers form a hierarchy
+that can be evaluated from top to bottom.
+
+- For everything else, evaluation starts at the first `facility` child and works downward.
+delayed evaluation equations are evaluated along with all other equations. There is no difference
+here.
```
ax3l, line 365: Let us write explicitly also:
- When are [inheritances](https://pals-project.readthedocs.io/en/latest/lattice-element-parameters.html#naming-and-inheriting-parameters) evaluated and can other elements be referenced inside `expr()`?
- What happens if I inherit from an element or element parameter that is a deferred expression?

source/fundamentals.md
ax3l: What I meant was: it will make it harder to implement PALS parser(s), if we desire to implement more than 1-2 reference implementations. Maybe we want to stay with 1-2 powerful, interpreter-like parsers and only the final expanded lattice is concrete and statically able to be validated with schemas.

DavidSagan: As mentioned in the meeting, lattice expansion is complicated so as a practical matter I don't see the utility in trying to keep things simple to accommodate more reference implementations. Another point is "expr()" is a very minor complication.

ax3l: I think we can keep it minor if we define (limit) the interaction (precedence and mixing) with other dynamic features of PALS well.

ax3l:
> A typical Cornell CESR lattice file has nearly 300 controllers for defining bumps, chromaticity knobs, etc. And CESR is small being only 800 m in circumference. None of these controllers involve your two categories. There are many use cases for expressions.

Sorry for the confusion, I was referring to the textbook "delayed expression" use cases in MAD that I found and see in lattice files.
I have no doubt about the need for controllers.

source/fundamentals.md
DavidSagan: This is a delicate question that needs to be discussed on its own.  I have created an Issue #253 and will open a PR when I have formulated something.

source/fundamentals.md
DavidSagan: I agree.

qianglbl (commented):
@DavidSagan 1)Line 253, "Multiple definitions of the same constant or variable are allowed with the restriction that the value is always the same." If the value is always the same, why need multiple definitions? 2)Line 301, "If there are {math}`N` lattice elements that match the name `B1.*`, {math}`N` different random numbers need to be generated for the calculation." It is not clear how this can be done with only one expression "value: 0.01 + 0.003 * random_gauss()".

source/fundamentals.md
```diff
@@ -213,27 +216,48 @@ PALS:
...
-      - const_c:: pi * const_a  # Can use expressions.
+      - const_c: pi * const_a   # Can use expressions.
+      ...
+  - variables:
+      - var_a: var_val_a        # Define var_a
       ...
 ```
+The difference between a constant and a variable is one of intent of the PALS file creator.
+That is, a constant is not supposed to be
+varied after a program reads a PALS file while a variable can vary. For example, a constant could
```
cemitch99:
```suggestion
varied after a program reads a PALS file, while a variable can vary. For example, a constant could
```

source/fundamentals.md
```diff
@@ -213,27 +216,48 @@ PALS:
...
+be the length of a magnet which will not change while a variable could be a magnet strength which
```
cemitch99:
```suggestion
be the length of a magnet which will not change, while a variable could be a magnet strength which
```

source/fundamentals.md
```diff
@@ -263,3 +287,84 @@ mass_of(A)               # Mass of particle A
...
+```
+This sets the length of the `cleo` element to `0.1*log(abs(b_var))` and the `Kn1` parameter
+to `3.74 * a_var`. 
+
+There are two types of expressions
+that are commonly called "immediate evaluation" and "delayed evaluation" expressions.
+Immediate evaluation expressions are expressions that, once a PALS file is read in by a program,
+are meant to be immediately evaluated and never reevaluated. That is, what is stored is the
+value of the expression and the expression used to compute the value can be ignored thereafter. 
+Delayed evaluation on the other hand are meant to be retained by the program and, if any
```
cemitch99:
```suggestion
Delayed evaluation expressions, on the other hand, are meant to be retained by the program and, if any
```

source/fundamentals.md
```diff
@@ -263,3 +287,84 @@ mass_of(A)               # Mass of particle A
...
+
+- A "merged" tree is constructed where any include files are merged into the root file tree.
+
+- The controllers are evaluated last as a group. Controllers with a `RELATIVE` `control_type` do
+not initially affect any parameters (it is only if the program reading the PALS file later varies a 
+RELATIVE controller variable that there is any effect). The ABSOLUTE controllers form a hierarchy
+that can be evaluated from top to bottom.
+
+- For everything else, evaluation starts at the first `facility` child and works downward.
+delayed evaluation equations are evaluated along with all other equations. There is no difference
```
cemitch99:
```suggestion
Delayed evaluation equations are evaluated along with all other equations. There is no difference
```

source/fundamentals.md
```diff
@@ -263,3 +287,84 @@ mass_of(A)               # Mass of particle A
...
+here.
+
+Note: A PALS parser will both compute parameter values and will preserve the underlying equations
+so a program that can use the equations will have them available but other programs can just
+use the values for lattice calculations. In any case, any program can do optimization on any set 
+of parameters for stuff like lattice design independent of any expressions.
```
cemitch99:
```suggestion
of parameters (for tasks such as lattice design) independent of any expressions.
```

cemitch99 (dismissed):

PR #245, opened 2026-07-15
Name matching

DavidSagan:
Up to now the standard has only discussed lattice element and element parameter name matching. This PR generalizes the discussion to name matching of, say, controllers using the same syntax as for lattice elements.

### DIFF
--- a/source/fundamentals.md
+++ b/source/fundamentals.md
@@ -125,47 +125,61 @@ to the following:
 (s:name.matching)=
 ## Name Matching
 
-PALS defines a standard syntax for name matching.
-For element parameters, the general syntax is
+Name matching is the process of finding the set of named constructs — lattice
+elements, element parameter groups, element parameters, constants, variables,
+etc. — that a given string is matched to. PALS defines a standard syntax for this.
+
+At its simplest, a string is matched against the `name` of a construct. For
+example, `qa.*` matches every element (or constant, variable, ...) whose name
+begins with `qa`. Names may be [PCRE2](https://www.pcre.org/) regular
+expressions. Regex matching is a whole-name (fully anchored) match — the pattern
+must match the entire name — so `qa.*` matches any name beginning with `qa`,
+while `qa` on its own matches only the exact name `qa`.
+
+Element parameters are matched by appending the parameter path to an element name
+match, using a single `>` as the separator:
 ```{code} text
-{beamline}>>{element-name}>{parameter-group}.{sub-group1}. ... .{sub-groupN}.{parameter} or
-{branch-name}>>{element-name}>{parameter-group}.{sub-group1}. ... .{sub-groupN}.{parameter}
-{kind}>>{element-name}>{parameter-group}.{sub-group1}. ... .{sub-groupN}.{parameter}
+{element-name}>{parameter-group}.{sub-group1}. ... .{sub-groupN}.{parameter}
 ```
 where
 ```{code} yaml
-{beamline}                      # Optional BeamLine name.
-{branch-name}                   # Optional Branch name.
-{kind}                          # Optional element kind name.
-{element-name}                  # Optional element name.
+{element-name}                  # Element name match, see below.
 {parameter-group}               # Parameter group name.
 {sub-group1}. ... .{sub-groupN} # Subgroups if they exist.
 {parameter}                     # Parameter name.
 ```
-Only `{beamline}`, `{branch-name},` and `{element-name}` use PCRE2 syntax. 
+The `{element-name}` component is any element name match from
+[Element Name Matching](#s:element.matching) — including its kind (`::`),
+BeamLine/Branch (`>>`), and Lattice (`>>>`) qualifiers. The parameter path
+following the `>` is, unlike element names, matched exactly rather than with
+PCRE2; the dots within it are therefore unambiguous, and the single `>`
+separator is distinct from the `>>` and `>>>` qualifiers that may appear inside
+`{element-name}`.
 
 Example:
 ```{code} yaml
 qa.*>MagneticMultipoleP.Ks2L
-Quadrupole>>qa.*>MagneticMultipoleP.Ks2L
+Quadrupole::qa.*>MagneticMultipoleP.Ks2L
 ```
-The first line will match the `Ks2L` component of all elements whose name begins with `qa`. Notice that
-since only `{beamline}` and `{element-name}` use PCRE2 syntax, the dot separating the parameter group
-and the parameter is unambiguous. The second line is like the first except only `Quadrupole` kind
-elements are matched to.
+The first line matches the `Ks2L` component of `MagneticMultipoleP` for all
+elements whose name begins with `qa`. The second is the same but restricted to
+elements of kind `Quadrupole`.
 
-Element parameter groups can similarly be matched to by using the above syntax without any
-`{parameter}` component. Similarly, Elements can be matched to if the `{parameter-group}`
-and everything following is not present. Example:
+A trailing part of the path may be dropped to match the enclosing construct
+instead of a parameter: omit `{parameter}` to match a parameter group, or omit
+the `>{parameter-group}...` entirely to match the element itself. Example:
 ```{code} yaml
-qa.*>MagneticMultipoleP    # Match to MagnetMultipoleP parameter group.
-qa.*                       # Match to element.
-Controller>>cd.t           # Match to all controllers whose four character name starts with `cd` and ends with `t`.
+qa.*>MagneticMultipoleP    # Match the MagneticMultipoleP parameter group.
+qa.*                       # Match the element.
+Controller::cd.t           # Match all Controllers whose four-character name starts with `cd` and ends with `t`.
 ```
+Constructs that are not element parameters — for example
+[constants and variables](#s:constants) — are matched directly by name, with no
+`>`-qualified parameter path.
 
 %---------------------------------------------------------------------------------------------------
 (s:specialvalues)=
-### Special Values
+## Special Values
 
 Special values used in this document are:
 
--- a/source/fundamentals.md
+++ b/source/fundamentals.md
@@ -123,23 +123,12 @@ to the following:
 
 %---------------------------------------------------------------------------------------------------
 (s:name.matching)=
+(s:parameter.matching)=
 ## Name Matching
 
 Name matching is the process of finding the set of named constructs — lattice
 elements, element parameter groups, element parameters, constants, variables,
 etc. — that a given string is matched to. PALS defines a standard syntax for this.
-
-Name matches may be restricted to a given element kind using the notation
-```{code} yaml
-{kind}::{name}
-```
-where `{kind}` is the element kind and `{name}` is the element name.
-Example:
-```{code} yaml
-Marker::bpm.
-```
-This will match to all `Marker` elements whose name is four characters starting with `bpm`
-(since a dot matches to any single character, see below).
 
 At its simplest, a string is matched against the `name` of a construct. For
 example, `qa.*` matches every element (or constant, variable, ...) whose name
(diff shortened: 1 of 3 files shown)


DavidSagan:
Draft. Will request reviews when ready.

cemitch99 (dismissed):

ax3l (commented):
This looks good to me. I asked my trusty Codex 5.6-Sol to review it too, and here are its findings:

  - [P2] Define `::` for non-element constructs. `Controller::cd.t`

    https://github.com/pals-project/pals/blob/e2fab3dd5b20342204640ebb500bce91b2b2d0d6/source/fundamentals.md#L168-L175
    uses `::` as a generic construct-kind selector, while the existing definition
    only permits `{element-kind}::{element-name}` for lattice elements. Add a
    normative `{kind}::{name}` rule for arbitrary constructs, including
    namespace/context behavior.

  - [P2] Support ungrouped element parameters. The new grammar

    https://github.com/pals-project/pals/blob/e2fab3dd5b20342204640ebb500bce91b2b2d0d6/source/fundamentals.md#L139-L150
    requires a parameter group, but PALS defines length, kind, `is_on`, etc.
    outside any group

    https://github.com/pals-project/pals/blob/9d4c3a4092e93baa1b632029a46a385a7441e98e/source/lattice-element-parameters.md#L33-L45
    Define an alternative such as `{element-name}>{parameter}`, otherwise
    `Q.*>length` has no specified interpretation.

  - [P2] Qualify the claim that variables never use `>`. Lines 176–178

    https://github.com/pals-project/pals/blob/e2fab3dd5b20342204640ebb500bce91b2b2d0d6/source/fundamentals.md#L176-L178
    say variables are matched directly, but related PR #237 defines controller
    variables as `{controller-name}>{variable-name}`

    https://github.com/pals-project/pals/blob/b101ef3b9c0f3b05cbdb85bdc3604268d45d031d/source/miscellaneous.md#L89-L101
    Restrict this statement to top-level variables and define matching for
    nested controller variables.

  - [P3] Preserve the old cross-reference target. Removing `s:parameter.matching`
    will break incoming references; PR #237 already contains two. Retain it as a
    compatibility alias or update dependent PRs before merging.

DavidSagan:
@ax3l Comments addressed. Please take a look.

qianglbl (commented):
@DavidSagan 1)Line 134 "example, `qa.*` matches every element (or constant, variable, ...) whose name begins with `qa`."  Why use qa.* instead of qa*? 2)Line 136 "Regex matching is a whole...." It is not clear why need "Regex" there. 3)Line 153 "(since a dot matches to any single character, see below)". I don't see the discussion of "dot match" below. Why use "dot" match to any single character instead of "?" for single character matching.

DavidSagan:
@qianglbl 
- Line 134: Using `?` only matches to zero or one characters. 
- Line 136: There are other kinds of matching than regex so using `regex` in the sentence reinforces the type of matching.
- Line 153: I took out "see below".
- `?` in regex is a "qualifier" that tells how many to match to. Not what to match. A web search will provide examples.

DavidSagan:
@qianglbl Perhaps you are thinking of unix file name matching?

qianglbl:
@DavidSagan Yes, it is.

ax3l:
Yes, David so far uses [PCRE2](https://www.pcre.org) (see PR details and prior merged conventions in PALS). Whatever we pick, I would stay with one regrex syntax for clarity.

source/fundamentals.md
```diff
@@ -121,9 +121,94 @@ to the following:
...
+(s:parameter.matching)=
+## Name Matching
+
+Name matching is the process of finding the set of named constructs — lattice
+elements, element parameter groups, element parameters, constants, variables,
+etc. — that a given string is matched to. PALS defines a standard syntax for this.
+
+At its simplest, a string is matched against the `name` of a construct. For
+example, `qa.*` matches every element (or constant, variable, ...) whose name
+begins with `qa`. Names may be [PCRE2](https://www.pcre.org/) regular
```
ax3l, line 135: One thing I realized: if any name might be a PCRE2 string and we have no indicator when that is the case, that means: we need to disallow all characters as names that are PCRE2 special meaning: 

```
  \  ^  $  .  [  ]  |  (  )  *  +  ?  {  }  #  /  -     and all whitespace
```

That leaves, from printable ASCII as allowed: `A-Z a-z 0-9` plus ``! " % & ' , : ; < = > @ _ ` ~``

We could write this list of forbidden characters down next to the reserved keywords where we already restrict names.

source/lattice-elements.md
ax3l, line 186: `>` is missing here in the precendence table and needs to be added, since it is added in this PR.

source/fundamentals.md
```diff
@@ -121,9 +121,94 @@ to the following:
...
+while `qa` on its own matches only the exact name `qa`.
+
+Name matches may be restricted to a given construct kind using the notation
+```{code} yaml
+{kind}::{name}
+```
+where `{kind}` is the construct kind — for example a lattice element kind such as
+`Marker`, or a non-element kind such as `Controller` — and `{name}` is the name
+match. The `{kind}` selector resolves within the namespace of the given kind, so
+the same `{name}` may match different constructs depending on the kind supplied.
```
ax3l: Claude found:

"Resolves within the namespace of the given kind" doesn't say what the namespaces are or what an unqualified name resolves against. This bites now: the merged Controllers section names variables `ps27>cur1`, lexically identical to `{element-name}>{parameter}`.

source/fundamentals.md
```diff
@@ -121,9 +121,94 @@ to the following:
...
+Marker::bpm.
+```
+This will match to all `Marker` elements whose name is four characters starting with `bpm`
+(since a dot matches to any single character).
+
+Element parameters are matched by appending the parameter path to an element name
+match, using a single `>` as the separator. A parameter that belongs to a
+[parameter group](#s:param.groups) is matched using the group path:
+```{code} text
+{element-name}>{parameter-group}.{sub-group1}. ... .{sub-groupN}.{parameter}
```
ax3l, line 157: Claude found:
*PCRE2 metacharacters collide with the separators*: `{2,3}` vs `,`, `(?:...)` vs `:`, `(?>...)/(?<n>...)` vs `>`, `(?#...)` vs `#`.

Pre-existing PALS issue, but this PR is what merges regex + separators into one general grammar, so it's the natural place for a longest-match tokenization rule or a stated limitation.

ax3l (commented):
This is pretty good, but we need to ensure we avoid issues with colliding PCRE2 special-meaning characters and PALS-reserved separators like `{}`.

source/fundamentals.md
DavidSagan: The characters used for a name is already restricted to `A-Z, a-z, 0-9, and _` (see the `Names` section) so no special characters can appear.

source/fundamentals.md
DavidSagan: I don't understand this. Can you provide an example?

source/lattice-elements.md
DavidSagan: Added.

source/fundamentals.md
DavidSagan: Sentence removed.

DavidSagan:
@ax3l Please respond to comments.

source/fundamentals.md
ax3l: Yay, good point - perfect!

source/fundamentals.md
ax3l: I think the problem is when to split on _separators_ like PALS keys `>>> >> :: > # : , &`.

A PRCE2 line that would match validly `Q12,Q123` could read: `Q\d{2,3}` and then being confused when to split in parsers...

**Claude's recommended fix**: state that _PALS separators_ `>>> >> :: > # : , &` are recognized only at **nesting depth zero**: _outside_ `(...)`, `[...]` and `{...}`.

DavidSagan: I think Claude is confused. PRCE2 is only applied to a name field. For example:
```
(q1.*>MagneticMultipole.Kn1 - ...)
```
In this example, PRCE2 is applied to the string `q1.*` and this is a valid construct.

Note: The above example evaluates to an array. Arrays are something that are not (yet) in the standard and would need to be discussed in a future PR.

ax3l: Yeah, I thought we already separated regex matching and PALS separator in another PR, let me search for evidence.

DavidSagan: In the `Element Name Matching` section:
```
Regex matching is applied to the lattice name, branch name, and element name separately
```

DavidSagan: And I have expanded the doc to make this clearer.

ax3l: I think the problem is not that PCRE2 is applied after we tokenize into structural groups, but that tokenization in groups itself is unclear.

The precedence list we defined in `lattice-elements.md` is a splitting rule, and that splitting has to happen before regex compilation in names. So valid names with regexes like these cannot unambiguously be split:

  | Match string | Legal PCRE2 | Splits into |
  |---|---|---|
  | `Q\d{2,3}` | matches `Q12`, `Q123` | `Q\d{2` + `3}` |
  | `(?:qa\|qb)1` | non-capturing group | `(?` + `qa\|qb)1` |
  | `(?>qa\|qab)1` | atomic group | name `(?` + param `qa\|qab)1` |

The problem is basically that `>`, `,` and `:` are PALS separators that also occur inside legal PCRE2 constructs, so an implementation's tokenizer cannot know where names (or names with regex) end and PALS tokens begin.

The first two are the most likely to occur in practice and have workarounds (`Q\d\d\d?`, `(qa|qb)1`). But they are perfectly valid PCRE2, so someone will eventually write one, and right now two conforming parsers can split it differently and get different lattices from the same file.

We could state the tokenization rule and document the how we like ambiguous cases to fail.

  > PALS separators are located by a left-to-right scan of the match string, performed before any
  regex compilation. Consequently, PCRE2 constructs containing a separator character, e.g., `{n,m}`,
  `{n,}`, `(?:`, `(?>`, `(?<name>`, and `(?#`, cannot be used in a name match.

Escaping does not help, e.g.,  `Q\d{2\,3}` compiles without error but silently stops being a quantifier (it becomes the literal text `{2,3}`).   The point is less about which constructs we lose than about implementations agreeing.

The alternative is the depth-zero rule as above, where we say PALS separators only exist outside of balanced `(...)`,  `[...]` and `{...}`. We can do full PCRE2 matching and force PALS implementations to track bracket depths/balancing before splitting into tokens.

I kinda prefer the parenthesis balancing rule. The formal phrasing could be:
  > **Tokenization.** A match string is split into PALS tokens before any regex is compiled. PALS
  > separators (`>>>`, `>>`, `::`, `>`, `#`, `:`, `,`, `&`) are recognized only at bracket depth
  > zero. Depth is computed in a single left-to-right scan in which:
  >
  > - a backslash escapes the next character, which is consumed without affecting depth (`\(`, `\,`);
  > - a `\Q` ... `\E` span is consumed without affecting depth;
  > - `(`, `[` and `{` increase depth; `)`, `]` and `}` decrease it;
  > - within a character class, no bracket other than the closing `]` is structural, and a `]`
  >   immediately after `[` or `[^` is a literal rather than the terminator.
  >
  > Because names are restricted to `A-Z a-z 0-9 _`, every bracket in a match string is regex
  > syntax. A match string whose brackets are unbalanced under this scan is invalid, even where
  > PCRE2 itself would accept them as literals. The comment construct `(?#...)` is not supported.

But it is exactly as tricky to implement in practice as it reads here.

ax3l: Probably disallowing a hole lot of valid PCRE2 constructs is the only way to keep this in check... What do you think @DavidSagan ?

ax3l: We could also simplify a bit more radically, and just replace PCRE2 regex in PALS with simple unix-like globbering as @qianglbl brought up elsewhere, say, we only allow `?` and `*` wildcards. Probably solves 95% of our needs as well and is so much simpler.

DavidSagan: @ax3l The other possibility is to say that if any of these more complicated constructs is used, the name field has to be enclosed in, say, backticks. Your other two suggestions also seem reasonable to me. Perhaps this question could be talked about at the next meeting.

qianglbl (approved):

PR #258, opened 2026-07-19
Route theme search box to Read the Docs server-side search

DavidSagan:
## Problem

The theme's search box (left sidebar + top-bar button) opens sphinx-book-theme's built-in **client-side** search. That search is page-granular and doesn't surface sub-section headings well — e.g. searching `include` never gives a good hit for the **Including Other Files Within PALS files** section (it ranks the page under its H1, "Fundamental Structure", and can't link to the section anchor).

Read the Docs already builds a much better **server-side** search that finds that section by name with the correct anchor — but out of the box it's only reachable from the RTD flyout menu and the `/` hotkey, which is easy to miss. RTD does **not** automatically override a theme's search box.

## Fix

Per RTD's [documented integration](https://docs.readthedocs.com/platform/stable/addons.html), the desired search input should dispatch the `readthedocs-search-show` event. This PR adds `source/_static/rtd-search-override.js`, which intercepts clicks on the theme's existing search buttons (sidebar + top bar) and opens the RTD server-side search instead — so the prominent box becomes the good search, with no new box added.

- `source/_static/rtd-search-override.js` — the override (guards on RTD being present).
- `source/conf.py` — enable `html_static_path = ['_static']` and register the JS via `html_js_files`.

The script guards on RTD Addons being present, so **local `make html` builds fall back to the theme's built-in search** with no behavior change.

## Verification

Verified locally:
- Full docs build succeeds.
- The JS is copied to `_static/` and referenced on every page.
- Both theme search buttons match the script's selector.
- On a local build the guard leaves the theme search untouched (no regression).

RTD's addon JS only runs on hosted builds, so the click-to-server-side-search behavior can only be confirmed on the **PR preview build**: open the left-sidebar search on the preview and confirm it now finds sections like *Including Other Files…* by name.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/source/_static/rtd-search-override.js
+++ b/source/_static/rtd-search-override.js
@@ -1,0 +1,54 @@
+// Route the sphinx-book-theme search box to Read the Docs' server-side search.
+//
+// By default the theme's search box (sidebar + top-bar button) opens the
+// theme's built-in client-side search, which is page-granular and does not
+// find sub-section headings well. Read the Docs ships a much better
+// server-side search, but out of the box it is only reachable from the RTD
+// flyout menu and the "/" hotkey. This script makes the existing, prominent
+// search box open the RTD server-side search instead.
+//
+// RTD Addons only run on the hosted readthedocs.io build. On a local
+// `make html` build the RTD elements are absent, so we detect that and leave
+// the theme's built-in search untouched (no-op fallback).
+
+(function () {
+  "use strict";
+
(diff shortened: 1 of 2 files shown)


DavidSagan:
Unfortunately the search index is not built for the PR so the new search cannot be tested until it is merged. I will merge and test and revert if needed.

DavidSagan:
@jlvay @cemitch99 @EZoni @ax3l The new search is definitely better than the old so I am leaving this in.

Issue #125, opened 2025-11-12
PALS Version Attribute


ax3l:
In openPMD, we have an important attribute that self-describes the version of the standard used in each file.
I would propose we do the same for PALS, increasing reproducibility and clarifying what to expect when opening a PALS file.

I propose to use semantic versioning for the standard.

ax3l:
Discussed in the meeting today:

Currently, we report on PALS in conference papers around twice a year. Starting to "tag" versions of PALS that we reference and creating automatic Zenodo DOIs would be a good practice now (cc @EZoni), also given that implementations need to start referencing what version of PALS they implement.

I try to avoid cal(ender)-ver, because it is a bit random to know what changed besides time (e.g., `26.06`).

I would propose for now to start with a [zero-ver](https://0ver.org/) schema:
- 0.1
- 0.2
- 0.3

all do not need to be backward compatible. This is probably my preferred option for now, and can always be forward-changed to [semver](https://semver.org/) or [cal-ver](https://calver.org/) later on.

@DavidSagan raised a compromise could be:
- 0.1.26.06  (the 0-ver 0.1 release tagged in June 2026)
- 0.2.26.12 (the 0.2 released tagged in Dec 2026)
- etc

Technically, the release date is always encoded in the git tag & release, and also in the Zenodo DOI, but it makes it visually very obvious that way, too. A downside I saw in the past is that people get easily confused if major/minor/patch versions are not consecutive.

EZoni:
Given that a few standard schemes already exist, such as the ones you mentioned (zero-ver, sem-ver, cal-ver), I would be in favor of using one of those, rather than a new custom one, such as the one proposed by David. As you mention, the release date is easily accessible even if it's not encoded in the version string. I'm fine with either zero-ver or sem-ver - I believe we use sem-ver for PALS-Python, where patches may occur more frequently.

ax3l:
We seem to like SemVer (maybe starting with Zero-Ver) so far, used in #260 for the `load` command. I agree this is a good choice.

PR #260, opened 2026-07-19
Load

DavidSagan:
Open PR #183 defines a `load` command meant to "allows to compose complex lattices over multiple PALS files, which otherwise are fully standalone and standard compliant files." This is definitely needed but there are practical problems with how the `load` command was defined. Specifically, the `load` command was not general enough and did not address issues like how to join machines. This PR defines a more general `load` command to address these issues. The PRs are incompatible so at most only one should be merged.


source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+        branches:
+          - ...
+```
+and a settings file could look like:
+```{code} yaml
+# Settings file: settings.pals.yaml
+PALS:
+  notes:
+    - "Settings for 12 mrad crossing angle, 0.23 m beta_y at IP."
+    - "This is a second note."
```
cemitch99:
```suggestion
    - "This is a second node."
```

cemitch99:
@David @ax3l What is the relationship between this PR and PR #183 ?

EZoni:
Given that #183 was labeled `schema: structural changes`, I will label this one the same way.

EZoni:
> @david @ax3l What is the relationship between this PR and PR #183 ?

This PR's description says

> This PR defines a more general `load` command to address these issues. The PRs are incompatible so at most only one should be merged.

which sounds to me like they are mutually exclusive and should be discussed together.

DavidSagan:
> which sounds to me like they are mutually exclusive and should be discussed together.

Correct. They are exclusive.

ax3l:
Hi y'all,

Thanks for advancing the `load` command @DavidSagan ! I have no strong feeling about my draft in #183 and we can continue here if you all like :)

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+  facility:
+    - sets:
+        - B35W>MagneticMultipole.Kn0L: 0.07
+        - ...
+```
+
+The rules for using `load` are as follows:
+- The `load` node must be a component of the `PALS` node. 
+- The top level of the joiner and loaded files must be `PALS`. 
+- A `SELF` designation in the `load` list indicates where the contents of
```
ax3l, line 196: `SELF`: good entry for a "Reserved Keywords" section (see: https://github.com/pals-project/pals/pull/282#discussion_r3706013149) at the end of `fundamentals.md`

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+  ```{code} yaml
+  notes:
+    - "Layout as of 03/12/2027"
+    - "Settings for 12 mrad crossing angle, 0.23 m beta_y at IP."
+    - "This is a second note."
+    - "Lattice with orbit correction for blown chopper at B34W."
+  ```
+- For Dict type subnodes of `PALS`, the combined dict will be the union of all the dict entries
+for all of the files. For any duplicate entries, the entry values must be the same and the combined Dict
+will discard the duplicates.
```
ax3l: What precedence do we want to define if the loaded keys do have different values? Error out or let one (well-defined) one win?

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+type subnodes, the combined list retains the order set by the `load` list and the order set within the files
+themselves. In the above example, the combined `notes` subnode will be:
+  ```{code} yaml
+  notes:
+    - "Layout as of 03/12/2027"
+    - "Settings for 12 mrad crossing angle, 0.23 m beta_y at IP."
+    - "This is a second note."
+    - "Lattice with orbit correction for blown chopper at B34W."
+  ```
+- For Dict type subnodes of `PALS`, the combined dict will be the union of all the dict entries
```
ax3l, line 222: Codex:

[P1]

> Define merging for scalar root nodes. The rules cover lists, dictionaries, and `version`, but `phase_space_coordinates` is also a valid scalar under `PALS`. Two loaded files can therefore produce an undefined result. Add a generic scalar rule—typically copy a sole value and require duplicate values to agree—which would also cover scalar extension keys.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+for all of the files. For any duplicate entries, the entry values must be the same and the combined Dict
+will discard the duplicates.
+- For the `version` subnode of `PALS`, if the version strings are not the same, the combined version
+will be a comma delimited list of the different versions with duplicates discarded.
+- A loaded file may itself contain a `load` node, so loading can be nested to any depth. It is
```
ax3l: Codex:

[P1]

> Define circular and repeated load behavior. “Nested to any depth” admits `A → B → A`, for which bottom-up expansion never terminates. Require cycle detection and declare cycles invalid. Also clarify diamonds—whether a shared file contributes once per load occurrence or once globally—because that changes ordered `facility` commands.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+  notes:
+    - "Layout as of 03/12/2027"
+    - "Settings for 12 mrad crossing angle, 0.23 m beta_y at IP."
+    - "This is a second note."
+    - "Lattice with orbit correction for blown chopper at B34W."
+  ```
+- For Dict type subnodes of `PALS`, the combined dict will be the union of all the dict entries
+for all of the files. For any duplicate entries, the entry values must be the same and the combined Dict
+will discard the duplicates.
+- For the `version` subnode of `PALS`, if the version strings are not the same, the combined version
```
ax3l: Codex:

[P1]

> Avoid synthesizing comma-delimited schema versions. `version` identifies the schema used by one file, but `"1.0, 2.0"` is not a single schema version and loses which content used which schema. The rule also omits valid `version: null` and missing-version behavior. Require compatible/equal versions, normalize before merging, or formally introduce a structured multi-version representation.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+PALS:
+  notes:
+    - "Lattice with orbit correction for blown chopper at B34W."
+  load:
+    - layout.pals.yaml
+    - settings.pals.yaml
+    - SELF
+  facility:
+    - sets:
+        - B35W>MagneticMultipole.Kn0L: 0.07
```
ax3l: Codex:

[P2]

> `MagneticMultipole.Kn0L` should be `MagneticMultipoleP.Kn0L`; `MagneticMultipole` is not the defined parameter-group name.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+        kind: Fork
+        to_line: AGS_ring           # Defined in AGS_ring.pals.yaml
+        destination_element: d24w   # AGS element at injection point
+
+    - AGS_inj_patch:
+        kind: Patch
+        ...
+
+    - AGS_inject:
+        kind: Beamline
```
ax3l: Codex:

[P2]

> `kind: Beamline` should use the defined, case-sensitive spelling `kind: BeamLine`.

source/fundamentals.md
ax3l, line 222: Codex:

[P1]

> Define whether dictionary merging is recursive. `extension_labels` contains nested `names`, `prefixes`, and `suffixes` dictionaries. Two files adding different entries under `names` either merge successfully or conflict depending on whether “union” is deep or shallow. Specify the depth and semantic equality rules.

source/fundamentals.md
ax3l: Codex:

[P1]

> State that `load` directives are consumed. The combined document should not retain its own or nested `load` nodes. Otherwise they could be processed twice, and nested relative paths would lose their original file context. Add an explicit rule that resolved `load` nodes do not appear in the combined `PALS` tree.

source/fundamentals.md
ax3l, line 196: Codex:

[P2]

> Restrict `SELF` to at most one occurrence. Multiple `SELF` entries currently have no defined validity or behavior and could duplicate the joiner’s notes, definitions, and ordered commands.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+    - SELF
+  facility:
+    - sets:
+        - B35W>MagneticMultipole.Kn0L: 0.07
+        - ...
+```
+
+The rules for using `load` are as follows:
+- The `load` node must be a component of the `PALS` node. 
+- The top level of the joiner and loaded files must be `PALS`.
```
ax3l: Codex:

[P2]

> Preserve permitted data outside the `PALS` node. The existing standard allows and ignores top-level siblings, while the new rule says the top level “must be `PALS`.” Rephrase this as “must contain a `PALS` node; only that node is merged” so otherwise-valid files remain loadable.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+for all of the files. For any duplicate entries, the entry values must be the same and the combined Dict
+will discard the duplicates.
+- For the `version` subnode of `PALS`, if the version strings are not the same, the combined version
+will be a comma delimited list of the different versions with duplicates discarded.
+- A loaded file may itself contain a `load` node, so loading can be nested to any depth. It is
+easiest to think of the process from the bottom up: each loaded file is fully combined into a single
+`PALS` file before it, in turn, is loaded into the joiner file that references it. There is no
+ambiguity as to what `SELF` refers to; a `SELF` designation always refers to the joiner file in
+which it appears.
+- As with `include`, a file path in a `load` list is interpreted relative to the location of the
```
ax3l, line 244: Codex:

[P2]

> Define failure semantics for unreadable loads. A missing, inaccessible, malformed, or non-PALS loaded file should explicitly make the joiner invalid rather than allowing silent omission. Recovery after reporting the error can remain implementation-defined.

source/fundamentals.md
ax3l:
```suggestion
        - B35W>MagneticMultipoleP.Kn0L: 0.07
```

source/fundamentals.md
ax3l:
```suggestion
        kind: BeamLine
```

source/fundamentals.md
ax3l: I agree, I think allowing mixed PALS versions will be very confusing and impossible to maintain well. Should error out cleanly if no match.

source/fundamentals.md
ax3l, line 222: Codex:

[P1]

> Do not select dictionary merging solely from YAML node shape. Mapping-valued root nodes can represent atomic objects, including constants defined directly under `PALS`. Field-wise merging two different constant definitions could create a definition that appeared in neither file and bypass the no-redefinition rule. Distinguish schema-declared mergeable dictionaries from atomic objects and extensions.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+
+The rules for using `load` are as follows:
+- The `load` node must be a component of the `PALS` node. 
+- The top level of the joiner and loaded files must be `PALS`. 
+- A `SELF` designation in the `load` list indicates where the contents of
+the joiner file, if there is anything else other than a `load` node,
+are to be combined with the contents of the loaded files. If not present, contents of the joiner
+file are to be combined at the end. In the above example, the `SELF` line is not needed since it
+is last on the `load` list.
+- Contents from the joiner and loaded files are combined `PALS` subnode by `PALS` subnode. For list
```
ax3l: Codex:

[P1]

> Resolve multiple `use` statements. Concatenating complete `facility` lists can retain a `use` statement from every loaded standalone machine. The existing standard only defines a single selector, so the chosen lattice becomes ambiguous. Specify whether loaded selectors are discarded, multiple selectors are invalid, or the joiner’s selector takes precedence.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+type subnodes, the combined list retains the order set by the `load` list and the order set within the files
```
ax3l: Codex:

[P1]

> Resolve loaded `expand_lattice` statements. Each standalone file may contain an expansion trigger. After concatenation there can be multiple triggers, although the construction algorithm describes splitting the facility around one. Define whether these directives are forbidden in loaded files, consumed before merging, or processed in a specified order after the full merge.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+will be a comma delimited list of the different versions with duplicates discarded.
+- A loaded file may itself contain a `load` node, so loading can be nested to any depth. It is
+easiest to think of the process from the bottom up: each loaded file is fully combined into a single
+`PALS` file before it, in turn, is loaded into the joiner file that references it. There is no
+ambiguity as to what `SELF` refers to; a `SELF` designation always refers to the joiner file in
+which it appears.
+- As with `include`, a file path in a `load` list is interpreted relative to the location of the
+file that contains the `load` node.
+
+Loading can also be useful in constructing "composite" accelerator complexes from individual machines.
```
ax3l, line 249: Codex:

[P1]

> Address global name collisions when joining machines. Two independently valid machines can both define elements such as `Q1`; concatenation then violates the existing prohibition on redefining elements. Since joining machines is a stated use case, provide namespacing/renaming or explicitly require globally coordinated names across every loaded file.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+PALS:
+  load:
+    - Booster_to_AGS_line.pals.yaml
+    - AGS_ring.pals.yaml
+
+
+  facility:
+    - AGS_fork:
+        kind: Fork
+        to_line: AGS_ring           # Defined in AGS_ring.pals.yaml
```
ax3l: Codex:

[P2]

> Put Fork parameters under `ForkP`. The composite example places `to_line` and `destination_element` directly on the `Fork`; both belong inside its `ForkP` parameter group.

source/fundamentals.md
```diff
@@ -14,6 +14,7 @@ PALS:
   reminders:        # [list] Optional reminder messages to be printed when file is read.
   extension_labels: # [Dict] Optional extensions to PALS that the standard shall ignore.
   facility:         # [list] lattice elements, beamlines, lattices, parameter set commands, etc.
+  load:             # [list] Files to load. See the "Load Files" section.
```
ax3l: Codex:

[P2]

> Complete the root-node inventory. The edited “Possible subnodes” block still omits `include` and `phase_space_coordinates`, even though both are valid direct children of `PALS`.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+files can be combined. A common use case is where one file defines the layout of a machine and another
+file defines the magnet and other settings for a given machine configuration. The final lattice
+is the union of these two files. This gives the flexibility where multiple settings files
+need only reference one layout file.
+
+While `load` and `include` both combine the contents of separate files, they differ in how the
+combination is done. An `include` inserts the contents of a file verbatim at the point where the
+`include` appears, and so can be used at any level of the information tree. A `load`, by contrast,
+merges whole files `PALS` subnode by `PALS` subnode at the `PALS` root level, following the rules given
+below. When both are present, `include` statements are resolved first so that each file is complete
```
ax3l, line 135: Codex:

[P1]

> Define when semantic processing occurs. All includes and loads must finish before executing `set`, selecting `use`, expanding lattices, resolving references, or performing semantic validation. The settings example targets `Q1a`, which exists only after merging the layout; processing each source independently would reject or ignore it.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+"Loading" files is similar to [including files](#s:includefiles) in that the contents of several
+files can be combined. A common use case is where one file defines the layout of a machine and another
+file defines the magnet and other settings for a given machine configuration. The final lattice
+is the union of these two files. This gives the flexibility where multiple settings files
+need only reference one layout file.
+
+While `load` and `include` both combine the contents of separate files, they differ in how the
+combination is done. An `include` inserts the contents of a file verbatim at the point where the
+`include` appears, and so can be used at any level of the information tree. A `load`, by contrast,
+merges whole files `PALS` subnode by `PALS` subnode at the `PALS` root level, following the rules given
```
ax3l, line 134: Codex:

[P1]

> Define the translator representation after loading. The existing standard requires “exact” and “expanded” representations, while this PR introduces a distinct combined tree. Specify whether exact output retains `load` directives, returns resolved content, or whether translators must expose a third combined representation.

source/fundamentals.md
ax3l, line 244: Codex:

[P2]

> Define path origin when an include introduces a `load`. Because includes run first and may appear anywhere, an included fragment can insert a `load` node into its parent. “The file that contains the load node” could then mean either the fragment or the including PALS file, producing different relative paths. Choose one origin or prohibit this construction.

source/fundamentals.md
```diff
@@ -110,6 +111,139 @@ Other file endings indicate non-PALS data.
...
+- The `load` node must be a component of the `PALS` node. 
+- The top level of the joiner and loaded files must be `PALS`. 
+- A `SELF` designation in the `load` list indicates where the contents of
+the joiner file, if there is anything else other than a `load` node,
+are to be combined with the contents of the loaded files. If not present, contents of the joiner
+file are to be combined at the end. In the above example, the `SELF` line is not needed since it
+is last on the `load` list.
+- Contents from the joiner and loaded files are combined `PALS` subnode by `PALS` subnode. For list
+type subnodes, the combined list retains the order set by the `load` list and the order set within the files
+themselves. In the above example, the combined `notes` subnode will be:
```
ax3l: Codex:

[P2]

> Extend the identical-constant exception to loaded files. Existing rules allow identical constant redefinitions specifically when they come from multiple includes. Whole-file loading can create the same situation, especially when standalone files repeat shared constants, but remains prohibited by the current wording.

source/fundamentals.md
ax3l: Codex:

[P2]

> Add `new_branch` to the composite Fork. Even after nesting the parameters under `ForkP`, the example claims the AGS ring is pulled in as a new branch but omits `new_branch`. With its default `null`, the standard requires the destination to belong to an already-existing branch.

source/fundamentals.md
ax3l: Yes, but I would say fixing this belongs in a separate PR.

ax3l:
I reviewed what Codex annotated and agree, those are pretty valid points.
Sometimes it has multiple issues with the same line, so we must be careful not to mark it resolved after we solved, e.g., 1 of 3 issues on the same line.

Structural features like these have a long rat-tail of corner cases, so I am not surprised it brings those up in that quantity. I think they are all solvable though.

source/fundamentals.md
DavidSagan: Next push will have a reserved words section.

source/fundamentals.md
DavidSagan: The use of "must" means this would be an error.

source/fundamentals.md
DavidSagan: Next push will say that scalar nodes must have the same value.

source/fundamentals.md
DavidSagan: I can see erroring if the versions are incompatible but do you really want to error for versions 1.2.1 and 1.2.5? I have added that major versions need to be the same

ax3l: That is reasonable. If we decide for semver we can def do that!

source/fundamentals.md
DavidSagan: Done on next push.

source/fundamentals.md
DavidSagan: Done in next push.

source/fundamentals.md
DavidSagan: Done on next push.

source/fundamentals.md
DavidSagan: Done in next push.

source/fundamentals.md
DavidSagan: Done in next push.

source/fundamentals.md
DavidSagan: done in next push.

source/fundamentals.md
DavidSagan: Done on next push.

source/fundamentals.md
DavidSagan: Done on next push.

source/fundamentals.md
DavidSagan: This has been reworded on the next push.

source/fundamentals.md
DavidSagan: This is not a problem since the last use node is the one that counts.

source/fundamentals.md
DavidSagan: Done in next push.

source/fundamentals.md
DavidSagan: done in next push.

source/fundamentals.md
DavidSagan: Right now that is an error and I am leaving that as such. I will leave it to a future PR if someone wants to do something like define namespaces.

source/fundamentals.md
DavidSagan: Fixed in next push.

source/fundamentals.md
DavidSagan: This concerns lattice expansion and subsequent parsing. This is for a different PR.

source/fundamentals.md
DavidSagan: “exact” and “expanded” concepts are being removed in PR #292

source/fundamentals.md
DavidSagan: Clarified in the next push.

source/fundamentals.md
DavidSagan: This can be the subject of a future PR but for this PR it will remain prohibited.

source/fundamentals.md
DavidSagan: This was changed by a previous PR and `new_branch` is not required.

DavidSagan:
@ax3l Ready for rereview.

ax3l:
Thanks! Can you merge `main` into it to resolve the conflict in `fundamentals.md` pls?

source/fundamentals.md
```diff
@@ -8,12 +8,15 @@
...
-  version:          # [string] Version of the PALS schema used in this file
-  authors:          # [list] Authors associated with this file
-  notes:            # [list] Optional notes of interest.
-  reminders:        # [list] Optional reminder messages to be printed when file is read.
-  extension_labels: # [Dict] Optional extensions to PALS that the standard shall ignore.
-  facility:         # [list] lattice elements, beamlines, lattices, parameter set commands, etc.
+  authors:                  # [list] Authors associated with this file
+  extension_labels:         # [Dict] Optional extensions to PALS that the standard shall ignore.
+  facility:                 # [list] lattice elements, beamlines, lattices, parameter set commands, etc.
+  include:                  # [
```
ax3l: comment is kaputt :)

```suggestion
  include:                  # [string] Optional literal insert of a file
```

DavidSagan:
@ax3l Done and ready for rereview.

source/fundamentals.md
```diff
@@ -112,6 +115,178 @@ Other file endings indicate non-PALS data.
...
+- A `SELF` designation in the `load` list indicates where the contents of
+the joiner file, if there is anything else other than a `load` node,
+are to be combined with the contents of the loaded files. If not present, contents of the joiner
+file are to be combined at the end. In the above example, the `SELF` line is not needed since it
+is last on the `load` list.
+- Contents from the joiner and loaded files are combined `PALS` subnode by `PALS` subnode. 
+- For the `version` subnode of `PALS`, if the version strings are not the same, the combined version
+will be a comma delimited list of the different versions with duplicates discarded. It is an
+error for the versions to be incompatible. That is, if the major version is different.
+For example, version `1.4.2` is incompatible with `2.0.1`.
```
ax3l, line 205: Semver is a good choice. To close out in docs separately for #125

ax3l:
Hi @DavidSagan, thanks a lot for those updates, this is great! I took a fresh look and think this is very understandable and clean now.

ax3l (approved):
Yay, thanks for the teamwork on `load`! 🎉

PR #267, opened 2026-07-20
The last element does not have to be a Marker element.

DavidSagan:
The last element does not have to be a Marker element.

### DIFF
--- a/source/parameters/aperture.md
+++ b/source/parameters/aperture.md
@@ -203,6 +203,8 @@ the four possible solutions that connect the two vertex points at the ends of th
 the arc used is the arc of minimal length such that the center of the ellipse is on the same side
 as the `center` point with respect to a line drawn through the two points.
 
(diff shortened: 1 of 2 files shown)


source/parameters/aperture.md
```diff
@@ -203,6 +203,8 @@ the four possible solutions that connect the two vertex points at the ends of th
 the arc used is the arc of minimal length such that the center of the ellipse is on the same side
 as the `center` point with respect to a line drawn through the two points.
 
+Positive radius means that the arc is convex and negative radius means that the arc is concave.
+
```
cemitch99, line 207: This is not part of this PR, but lies in #266

DavidSagan: My mistake. I will fix at merge time if you approve this.

cemitch99 (approved):

PR #268, opened 2026-07-20
More detail about how lattice expansion works.

DavidSagan:
More detail about how lattice expansion works.


source/lattice-construction.md
```diff
@@ -49,25 +52,101 @@ element, this element must be dropped from the branch line.
...
+          kind: Quadrupole
+
+      - set:
+          parameter: Q.*>MagneticMultipoleP.Kn0
+          value: PARAMETER + 0.02
+
+      - Q2:                       # Defined after the set
+          kind: Quadrupole
+  ```
+  In this case since `Q2` is defined after the set, its MagneticMultipoleP.Kn0 value is not affected
```
cemitch99:
```suggestion
  In this case since `Q2` is defined after the set, its `MagneticMultipoleP.Kn0` value is not affected
```

cemitch99 (approved):

PR #282, opened 2026-07-28
Define `facility` for branches.

DavidSagan:
Define `facility` for branches.

### DIFF
--- a/source/lattice-elements.md
+++ b/source/lattice-elements.md
@@ -154,7 +154,7 @@ PALS:
     - mylat:
         kind: Lattice
         branches:
-          - mybranch              # mybranch branch will have a q1 element
+          - mybranch:             # mybranch branch will have a q1 element
               inherit: myline     #   once the lattice is expanded.
 
     - expand_lattice
--- a/source/lattice-elements.md
+++ b/source/lattice-elements.md
@@ -137,8 +137,8 @@ name with `">>>"` as a separator. There are several permutations where `>>` and 
 {branch-name}>>{element-name}
 ```
 
-If `NULL` is used as a branch name, the specified element is searched for outside of any lattices
-or beamlines.
+If `FACILITY` is used as a branch name, the specified element is searched for under the `facility`
+node. This will match to element definitions. For example:
 For example
 ```{code} yaml
 PALS:
@@ -160,14 +160,15 @@ PALS:
     - expand_lattice
 
     - sets:
-        - NULL>>q1>length: 0.4       # Matches to q1 definition in facility 
+        - FACILITY>>q1>length: 0.4       # Matches to q1 definition in facility 
         - mybranch>>q1>length: 0.5   # Matches to q1 in mybranch branch
         - q1>length: 0.6             # Matches to all three q1 instances
 ```
 
-Here `NULL>>q1` would match to the definition of `q1` outside of `myline` and
+Here `FACILITY>>q1` would match to the definition of `q1` outside of `myline` and
 `mybranch` while `mybranch>>q1` would match the instance of `q1` inside `mybranch`.
-The simple `q1` would match to all three instances.
+The simple `q1` would match to all three instances. It is not allowed to use the word
+`FACILITY` as a name for any lattice element, `BeamLine`, or `Lattice`. 
 
 Regular expressions can be used. 
 Regular expressions must conform to the [PCRE2](https://www.pcre.org/) standard. 
--- a/source/lattice-elements.md
+++ b/source/lattice-elements.md
@@ -158,15 +158,20 @@ PALS:
               inherit: myline     #   once the lattice is expanded.
 
     - sets:
-        - facility>>q1>length: 0.4       # Matches to q1 definition in facility 
-        - mybranch>>q1>length: 0.5   # Matches to q1 in mybranch branch
-        - q1>length: 0.6             # Matches to all three q1 instances
+        - facility>>q1>length: 0.4           # Matches to q1 definition in facility
+        - myline>>q1>ApertureP.x_min: 0.5    # Matches to q1 in myline.
+        - q1>ApertureP.y_max: 0.6            # Matches to both q1 instances.
 ```
 
-Here `facility>>q1` would match to the definition of `q1` outside of `myline` and
-`mybranch` while `mybranch>>q1` would match the instance of `q1` inside `mybranch`.
-The simple `q1` would match to all three instances. It is not allowed to use the word
-`facility` as a name for any lattice element, `BeamLine`, or `Lattice`. 
+Here `facility>>q1` would match to the definition of `q1` in the `facility` list.
+while `myline>>q1` would match the instance of `q1` inside `myline`.
+The simple `q1` would match to both of these instances. Notice that 
+since lattice expansion has not yet been done, nothing within the `mylat` lattice
+is immediately affected. However, when lattice expansion is done, 
+the `mylat` lattice will have an instance of `q1` whose parameter values will be affected
+by the prior sets before lattice expansion.
+Note: It is not allowed to use the word
+`facility` as a name for any lattice element, branch, `BeamLine`, or `Lattice`. 
 
 Regular expressions can be used. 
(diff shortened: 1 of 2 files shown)


source/lattice-elements.md
```diff
@@ -137,6 +137,22 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
...
+PALS:
+  facility:
+    - q1:
+        kind: Quadrupole
+
+    - myline:
+        kind: BeamLine
+        line:
+          - q1   
+```
```
EZoni: I don't see any `NULL` in this example. I feel like the actual example should be expanded and modified to be more self-explanatory with regard to what it is meant to exemplify.

Also, I think it should be highlighted as `yaml`, not `text`:

```suggestion
```{code} yaml
PALS:
  facility:
    - q1:
        kind: Quadrupole

    - myline:
        kind: BeamLine
        line:
          - q1   
```
```

Comment (DavidSagan) [reply source/lattice-elements.md]: Look at revised doc.

**Comment by DavidSagan** [comment]
@EZoni Ready for rereview.

File: source/lattice-elements.md [hunk 2]
```diff
@@ -137,6 +137,38 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
 {branch-name}>>{element-name}
 ```
 
+If `NULL` is used as a branch name, the specified element is searched for outside of any lattices
+or beanlines
```
ax3l:
```suggestion
or beamlines.
```

source/lattice-elements.md
```diff
@@ -137,6 +137,38 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
...
+    - myline:
+        kind: BeamLine
+        line:
+          - q1                    # q1 in the beamline myline
+
+    - mylat:
+        kind: Lattice
+        branches:
+          - mybranch              # mybranch branch will have a q1 element
+              inherit: myline     #   once the lattice is expanded.
```
ax3l: Broken YAML syntax:
```suggestion
          - mybranch:             # mybranch branch will have a q1 element
              inherit: myline     #   once the lattice is expanded.
```

source/lattice-elements.md
```diff
@@ -137,6 +137,38 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
 {branch-name}>>{element-name}
 ```
 
+If `NULL` is used as a branch name, the specified element is searched for outside of any lattices
```
ax3l: Since `NULL` is not a protected keyword in PALS, it can collide with a user lattice or beamline that they legitimately named `NULL`.

source/lattice-elements.md
```diff
@@ -137,6 +137,38 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
...
+        branches:
+          - mybranch              # mybranch branch will have a q1 element
+              inherit: myline     #   once the lattice is expanded.
+
+    - expand_lattice
+
+    - sets:
+        NULL>>q1.length: 0.4       # Matches to q1 definition block 
+        mybranch>>q1.length: 0.5   # Matches to q1 in mybranch branch
+        q1.length: 0.6             # Matches to all three q1 instances.
```
ax3l: inconsistent with definition of `set` in `source/miscellaneous.md` and `source/lattice-element-parameters.md`

You probably meant to write:
```suggestion
    - sets:
        - NULL>>q1>length: 0.4       # Matches to q1 definition in facility 
        - mybranch>>q1>length: 0.5   # Matches to q1 in mybranch branch
        - q1>length: 0.6             # Matches to all three q1 instances
```

source/lattice-elements.md
ax3l: I think `NULL` is not self-describing enough and thus potentially confusing.

Defining a reserved keyword like `FACILITY` would be way more explicit, communicating the "directly under the `facility`" intention better.

source/lattice-elements.md
ax3l: The last assignment matches all three earlier elements and overwrites the first two values, weakening the example. Separate examples or manipulation of a different element would demonstrate the selectors more clearly.

ax3l (changes requested):

source/lattice-elements.md
DavidSagan: Can you suggest a better alternative?

DavidSagan: Why not make `NULL` a protected keyword?

DavidSagan: OK switched to `FACILITY`.

DavidSagan: Moot since  switched to `FACILITY`.

source/lattice-elements.md
DavidSagan: I like the compactness of this example. We can ask if anyone would be confused by this and switch if there looks to be a problem.

qianglbl (commented):
@DavidSagan Line 141, "For example: ", Line 142, "For example" used twice. 
I see that "set: " can set "q1" value at multiple locations. What would happen to the q1 value outside the branch if one sets its value inside the branch (Line 162)?

source/lattice-elements.md
```diff
@@ -137,6 +137,37 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
...
+    - mylat:
+        kind: Lattice
+        branches:
+          - mybranch:             # mybranch branch will have a q1 element
+              inherit: myline     #   once the lattice is expanded.
+
+    - sets:
+        - facility>>q1>length: 0.4       # Matches to q1 definition in facility 
+        - mybranch>>q1>length: 0.5   # Matches to q1 in mybranch branch
+        - q1>length: 0.6             # Matches to all three q1 instances
```
ax3l: Formatting of the first comment:
```suggestion
        - facility>>q1>length: 0.4   # Matches to q1 definition in facility 
        - mybranch>>q1>length: 0.5   # Matches to q1 in mybranch branch
        - q1>length: 0.6             # Matches to all three q1 instances
```

source/lattice-elements.md
```diff
@@ -137,6 +137,37 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
...
+```
+
+Here `facility>>q1` would match to the definition of `q1` outside of `myline` and
+`mybranch` while `mybranch>>q1` would match the instance of `q1` inside `mybranch`.
+The simple `q1` would match to all three instances. It is not allowed to use the word
+`facility` as a name for any lattice element, `BeamLine`, or `Lattice`.
```
ax3l:
```suggestion
In the last line, the simple `q1` would match to all three instances and thus overwrites the two earlier `set` lines on `q1` elements.
It is not allowed to use the word `facility` as a name for any lattice element, `BeamLine`, or `Lattice`. 
```

To avoid confusion as we explain the example.

I think if we inverted the order of `set` lines:
- set all `q1`
- set `mybranch`
- set `facility`

This would be a bit more sensible (no "unused" lines in the example), but I am fine either way.

> It is not allowed to use the word `facility` as a name for any lattice element

I would probably add a new "Reserved Keywords" section at the end of `fundamentals.md` where we can add more revered keywords later on that users must not use for custom names and attributes.

ax3l (commented):
LGTM, small cleanup suggestions inline

source/lattice-elements.md
DavidSagan: Changed to set different parameters.

source/lattice-elements.md
DavidSagan: This is already changed.

source/lattice-elements.md
DavidSagan: This is changed.

DavidSagan:
Converting to draft for now since I want to review the PR.

source/lattice-elements.md
```diff
@@ -137,6 +137,42 @@ name with `">>>"` as a separator. There are several permutations where `>>` and
 {branch-name}>>{element-name}
 ```
 
+If `facility` is used as a branch name, the specified element is searched for under the `facility`
+node. This will match to element definitions. For example:
+For example
```
cemitch99, line 143:
```suggestion
```

DavidSagan:
PR is moot.

PR #285, opened 2026-07-30
Add a validated examples and unit-test corpus

ax3l:
Collect the valid PALS lattices from the [pals-cpp](https://github.com/pals-project/pals-cpp/tree/8d5cb48789fe86805d331e8ac67700953a83141f/tests) examples and unit tests into `examples/`:
two demonstrative machines (`machine/`, `modular_machine/`) and a feature-organized unit-test corpus (`unit_tests/`) covering the document root, element bookkeeping, beamlines, multipass, forks, expressions, controllers, sets, and `load`/`include` file families. Each file states in its header what it demonstrates and the `pals-cpp` test it comes from.

Every self-contained file expands without problems with `pals-cpp`'s `parse_and_expand_PALS`, members of load families are expanded through their `joiner.pals.yaml`.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,25 @@
+# PALS Examples
+
+Example lattices for the Particle Accelerator Language Standard, as physics examples (suitable for integration tests) and small unit tests.
+
+## Layout
+
+| Path | Contents |
+| --- | --- |
+| `fodo.pals.yaml` | The introductory example: elements, a FODO cell, `inherit`, `repeat`, a lattice, and `use`. |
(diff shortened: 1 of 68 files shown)


DavidSagan (approved):

ax3l:
Thanks for designing these, @DavidSagan !
Copied 1:1 so far.

Next steps after this PR:
- I will open a companion PR to `pals-cpp`, that will clone those and test them, so we can spot drifts between spec and impls. We will use the same examples in `pals-python` for testing.
- I will open a follow-up PR that updates the fodo example and other tests, checking if there is some drift between pals standard and test/examples already.

jlvay (approved):

DavidSagan:
Just be aware that the tests assume context from existing PRs will become part of PALS.

ax3l:
Makes sense, yes in those PRs we can update the test files here as the PRs change before merge.

PR #289, opened 2026-08-03
Document that `PARAMETER` can have varying values when applied to a set of parameters.

DavidSagan:
Document that `PARAMETER` can have varying values when applied to a set of parameters.

### DIFF
--- a/source/miscellaneous.md
+++ b/source/miscellaneous.md
@@ -30,8 +30,7 @@ Example:
 ```
 In this example, the `BendP.e1` parameter of all elements whose name begins with `B1` is modified.
 This includes element `B1a`. Notice that when a set of parameters is being changed, as in this example,
-the value of `PARAMETER` used in computation of a given parameter in the set will be the value 
-of that parameter. Thus the value of `PARAMETER` can be different for different parameters in the set.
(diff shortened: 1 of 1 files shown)


source/miscellaneous.md
```diff
@@ -29,7 +29,9 @@ Example:
     value: 2*PARAMETER + atan(SELF.BendP.g_ref)
 ```
 In this example, the `BendP.e1` parameter of all elements whose name begins with `B1` is modified.
-This includes element `B1a`. 
+This includes element `B1a`. Notice that when a set of parameters is being changed, as in this example,
+the value of `PARAMETER` used in computation of a given parameter in the set will be the value 
+of that parameter. Thus the value of `PARAMETER` can be different for different parameters in the set.
```
cemitch99: This description is difficult for a human to parse, and may cause more confusion than clarification.  Maybe it would help to make this more concise, e.g.  "Notice that when a set of parameters is being changed, as in this example, the value of `PARAMETER` can be different for different parameters in the set."

DavidSagan: Fixed.

cemitch99 (approved):

PR #293, opened 2026-08-06
Remove `expand_lattice` command and introduce `post_expansion`.

DavidSagan:
## Summary

The `post_expansion` construct simplifies the logic of lattice expansion so that operations like `load` are less convoluted. 

## To do

In follow-up PRs:
- https://github.com/pals-project/pals/pull/293#discussion_r3753538999: In post_expansion, does a bare `q1>...` match all expanded instances named `q1` across all lattices and branches?
- https://github.com/pals-project/pals/pull/293#discussion_r3753558565: Where do we explain what a new element definition means after expansion has completed? If there is no insertion/superposition mechanism, such a definition is inert and we should point that out, no?


source/fundamentals.md
```diff
@@ -54,6 +55,48 @@ to be outside of the PALS standard and will be ignored by a PALS parser.
...
+              repeat: 3
+
+    - lat:
+        kind: Lattice
+        branches: bline
+
+  post_expansion:
+    - set:
+        parameter: q1>MagneticMultipoleP.Kn1L
+        value: PARAMETER * (1 + 1e-4*random_gauss())
```
EZoni, line 93: In `post_expansion`, does a bare `q1>...` match all expanded instances named `q1` across all lattices and branches?

source/lattice-elements.md
```diff
@@ -39,7 +39,8 @@ crab1:
     voltage: 1.0e6
 ```
 
-Lattice element definitions may only be placed as a child node of the [`facility`](#s:palsroot) node
+Lattice element definitions may only be placed as a child node of the [`facility`](#s:palsroot)
+or [`post_expansion`](#s:palsroot) nodes,
```
EZoni, line 43: Where do we explain what a new element definition means after expansion has completed? If there is no insertion/superposition mechanism, such a definition is inert and we should point that out, no?

source/fundamentals.md
EZoni, line 1: The new text in `source/fundamentals.md` says the first "expansion" phase "basically involves constructing the ordered lists of lattice elements contained in all the branches" — that is branch expansion, not lattice expansion, according to the definitions in source/definitions.md. Meanwhile the step list in `lattice-construction.md` puts reference-parameter calculation before the `post_expansion` step, so phase one is actually full lattice expansion, if I'm not mistaken. Should these be reworded so that the two files agree?

source/lattice-construction.md
EZoni, line 1: Do we need to say how multiple `facility` and `post_expansion` nodes from different files are concatenated and in which order?

source/fundamentals.md
```diff
@@ -13,7 +13,8 @@ PALS:
   notes:            # [list] Optional notes of interest.
   reminders:        # [list] Optional reminder messages to be printed when file is read.
   extension_labels: # [Dict] Optional extensions to PALS that the standard shall ignore.
-  facility:         # [list] lattice elements, beamlines, lattices, parameter set commands, etc.
+  facility:         # [list] Lattice elements, beamlines, lattices, parameter set commands, etc.
+  post_expansion:   # [list] Set commands, etc. for post-expansion setup.
```
EZoni: Could we replace "etc." with an enumeration of what is and is not permitted here?

source/fundamentals.md
```diff
@@ -54,6 +55,48 @@ to be outside of the PALS standard and will be ignored by a PALS parser.
...
+
+The [lattice expansion](#s:lattice.expand) performed by a PALS parser can be divided into
+two phases. The first "expansion" phase basically involves constructing the ordered lists 
+of lattice  elements contained in all the branches of a lattice. The second "post-expansion"
+phase basically involves making modifications to the expanded lattice. 
+Essentially, the information needed to to the expansion phase is put in the `facility` sub-node
+of `PALS` and the information for the post-expansion work is put in the `post_expansion` sub-node.
+Example:
+```{code} yaml
+PALS
```
EZoni:
```suggestion
PALS:
```

source/fundamentals.md
```diff
@@ -54,6 +55,48 @@ to be outside of the PALS standard and will be ignored by a PALS parser.
...
 
+%---------------------------------------------------------------------------------------------------
+(s:facility)=
+## `facility` and `post_expansion`
+
+The [lattice expansion](#s:lattice.expand) performed by a PALS parser can be divided into
+two phases. The first "expansion" phase basically involves constructing the ordered lists 
+of lattice  elements contained in all the branches of a lattice. The second "post-expansion"
+phase basically involves making modifications to the expanded lattice. 
+Essentially, the information needed to to the expansion phase is put in the `facility` sub-node
```
EZoni:
```suggestion
Essentially, the information needed to do the expansion phase is put in the `facility` sub-node
```

source/fundamentals.md
```diff
@@ -54,6 +55,48 @@ to be outside of the PALS standard and will be ignored by a PALS parser.
...
+    - set:
+        parameter: q1>MagneticMultipoleP.Kn1L
+        value: PARAMETER * (1 + 1e-4*random_gauss())
+```
+In this example, the expanded lattice has three elements named `q1`. 
+To add a random error to each of these, the lattice has to be expanded before a `set` command is used,
+and this is accomplished by placing a `set` command under the `post_expansion` node.
+If the `set` command had appeared in the `facility` node, and not in the `post_expansion` node, then
+when the set command is executed, the set would be applied to the value of `Kn1L` in the definition
+of `q1`. In this case, result would be that all three of the `q1` elements in the expanded lattice would
```
EZoni:
```suggestion
of `q1`. In this case, the result would be that all three of the `q1` elements in the expanded lattice would
```

source/fundamentals.md
```diff
@@ -54,6 +55,48 @@ to be outside of the PALS standard and will be ignored by a PALS parser.
 PALS file `authors` are optional, but recommended to enable data provenance and contacts.
 Per author, the `name` is required; the `orcid`, `affiliation` and `email` fields are optional.
 
+%---------------------------------------------------------------------------------------------------
+(s:facility)=
+## `facility` and `post_expansion`
+
+The [lattice expansion](#s:lattice.expand) performed by a PALS parser can be divided into
+two phases. The first "expansion" phase basically involves constructing the ordered lists 
+of lattice  elements contained in all the branches of a lattice. The second "post-expansion"
```
EZoni:
```suggestion
of lattice elements contained in all the branches of a lattice. The second "post-expansion"
```

source/fundamentals.md
EZoni, line 93: This was `parameter` (lowercase) in the old code. What is the correct syntax?

source/lattice-elements.md
EZoni, line 43: Is it correct that both `facility` and `post_expansion` point to `#s:palsroot` here?

source/fundamentals.md
DavidSagan: I am saving this question for a later PR. Essentially what I will propose is that a plain `q1` refers to the expanded lattice or any element defined in `post_expansion`

source/lattice-elements.md
DavidSagan: Insertion and superposition will be needed post expansion so a new element definition will make sense. This will need a future PR.

source/lattice-construction.md
DavidSagan: That documentation will be in the `load` PR.

source/fundamentals.md
DavidSagan: I have reworded the section in fundamentals.md to make this clear.

source/fundamentals.md
DavidSagan: Fixed.

source/fundamentals.md
DavidSagan: This paragraph has been modified.

source/fundamentals.md
DavidSagan: Changed.

source/fundamentals.md
DavidSagan: This paragraph has been reworded.

source/fundamentals.md
DavidSagan: Upper case.

source/lattice-elements.md
DavidSagan: Yes.

source/fundamentals.md
DavidSagan: I put in something.

qianglbl (commented):
@DavidSagan "...set of node types..." Is the "node" defined somewhere? Does it mean an entry or an element or a line under the list?

DavidSagan:
@qianglbl According to the official [YAML Specification](https://yaml.org/spec/1.2.2/), a YAML node is a single structural element in a representation graph or serialization tree. It represents a native data structure and is classified into three structural kinds: scalar, sequence and mapping.

qianglbl:
@DavidSagan Thanks. It is clear to me now.

DavidSagan:
> @DavidSagan Thanks. It is clear to me now.

Great! Please approve PR if everything looks good.

qianglbl (dismissed):

source/fundamentals.md
EZoni: Unresolving this conversation just so it remains visible, since it refers to a fix that will be addressed (hence likely referenced through link) in a future PR.

source/lattice-elements.md
EZoni: Same as https://github.com/pals-project/pals/pull/293#discussion_r389665300: Unresolving this conversation just so it remains visible, since it refers to a fix that will be addressed (hence likely referenced through link) in a future PR.

cemitch99 (dismissed):

source/fundamentals.md
```diff
@@ -15,11 +15,14 @@ PALS:
...
   notes:                    # [list] Optional notes of interest.
   phase_space_coordinates:  # [enum] Type of phase space coordinates 
+  post_expansion:           # [list] Post-expansion setup. 
   reminders:                # [list] Optional reminder messages to be printed when file is read.
   version:                  # [string] Version of the PALS schema used in this file
 ```
 The difference between `notes` and `reminders` is that reminder messages are meant to be 
 printed (or otherwise communicated to the user) every time the file is read.
+The set of node types that can be used under the `post_expansion` node is the same as the
+`facility` node except for lattice and `use` nodes.
```
EZoni: Doesn't this belong better to the section "## `facility` and `post_expansion`" below, since that's where we tell the reader what the two nodes are for?

source/fundamentals.md
```diff
@@ -15,11 +15,14 @@ PALS:
   load:                     # [list] Files to load. See the "Load Files" section.
   notes:                    # [list] Optional notes of interest.
   phase_space_coordinates:  # [enum] Type of phase space coordinates 
+  post_expansion:           # [list] Post-expansion setup.
```
EZoni: Is `post_expansion` also optional? If so, maybe we should say it explicitly:

```suggestion
  post_expansion:           # [list] Optional post-expansion setup. 
```

source/fundamentals.md
DavidSagan: Moved in next push.

source/fundamentals.md
DavidSagan: Done in next push.

DavidSagan:
@EZoni Changes in. Ready for rereview.

EZoni:
It seems like there are conflicts to resolve in `source/definitions.md`.

DavidSagan:
@EZoni Conflict resolved.

qianglbl (approved):

cemitch99 (approved):

PR #294, opened 2026-08-23
New `Pull Request Discussion` section in Governance chapter.

DavidSagan:
New `Pull Request Discussion` section in Governance chapter.

### DIFF
--- a/source/governance.md
+++ b/source/governance.md
@@ -116,12 +116,14 @@ Anyone with an interest is invited to join the discussion of any Pull Request (P
 So that the standard can evolve at a reasonable pace, anyone who raises an issue in a PR
 discussion is encouraged to respond promptly to comments on the issue.
 If that is not possible, the person should post a note saying when they expect
-they can respond, and the discussion should wait for them accordingly.
+they can respond, and the decision to merge the PR should wait for them accordingly.
 
-If someone who raised an issue does not respond in any way within two weeks of the last
-message in that discussion thread, the decision of whether to merge the PR shall proceed
-without further delay. That is, the unanswered issue is treated as resolved and is not
-grounds for postponing the merge decision to allow more time for a response.
+If an issue has been raised, and if 
+1) all the comments supporting the issue have been responded to, and
+2) over two weeks has passed since the last comment on this issue has been posted, and 
+3) no one has asked for further time to respond, 
+then for the purposes of deciding whether to merge the PR, the issue is to be treated as resolved. 
+That is, the merge decision is not to be postponed to allow more time for responses.
 Notice that this two week limit does not apply to cases where a comment raises an issue and
 that comment has not been replied to.
 
--- a/source/governance.md
+++ b/source/governance.md
@@ -115,8 +115,7 @@ Anyone with an interest is invited to join the discussion of any Pull Request (P
 
 So that the standard can evolve at a reasonable pace, anyone who raises an issue in a PR
 discussion is encouraged to respond promptly to comments on the issue.
-It is recognized that vacations, scheduling conflicts, and the like, can make a timely
-response impossible. In that case, the person should post a note saying when they expect
+If that is not possible, the person should post a note saying when they expect
 they can respond, and the discussion should wait for them accordingly.
 
 If someone who raised an issue does not respond in any way within two weeks of the last
--- a/source/governance.md
+++ b/source/governance.md
@@ -122,6 +122,8 @@ If someone who raised an issue does not respond in any way within two weeks of t
 message in that discussion thread, the decision of whether to merge the PR shall proceed
 without further delay. That is, the unanswered issue is treated as resolved and is not
 grounds for postponing the merge decision to allow more time for a response.
+Notice that this two week limit does not apply to cases where a comment raises an issue and
+that comment has not been replied to.
 
 ## Former Members
 
--- a/source/governance.md
+++ b/source/governance.md
@@ -97,7 +97,7 @@ Contributions can be in many forms and not all need to be code contributions.
 Examples include contributing to pull requests, issues, user discussions, documentation, tutorials, etc.
 Contributors can participate in developer meetings and weigh in on discussions.
 Contributors can "triage" (add labels) to pull requests, issues, and GitHub discussion pages.
-Contributors can comment and review PRs (but not merge).
+Contributors can review PRs (but not merge).
 
 ### Decision Process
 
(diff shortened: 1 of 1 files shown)


cemitch99:
Suspect this PR would benefit from discussion with the technical committee or the larger group.

source/governance.md
```diff
@@ -109,10 +109,26 @@ Appointed after contributing to PALS (see above) by any member of the TC.
...
 
+## Pull Request Discussion
+
+Anyone with an interest is invited to join the discussion of any Pull Request (PR).
+
+So that the standard can evolve at a reasonable pace, anyone who raises an issue in a PR
+discussion is encouraged to respond promptly to comments on the issue.
+It is recognized that vacations, scheduling conflicts, and the like, can make a timely
+response impossible. In that case, the person should post a note saying when they expect
+they can respond, and the discussion should wait for them accordingly.
```
EZoni:
```suggestion
If that is not possible, the person should post a note saying when they expect
they can respond, and the discussion should wait for them accordingly.
```

source/governance.md
```diff
@@ -109,10 +109,25 @@ Appointed after contributing to PALS (see above) by any member of the TC.
...
+## Pull Request Discussion
+
+Anyone with an interest is invited to join the discussion of any Pull Request (PR).
+
+So that the standard can evolve at a reasonable pace, anyone who raises an issue in a PR
+discussion is encouraged to respond promptly to comments on the issue.
+If that is not possible, the person should post a note saying when they expect
+they can respond, and the discussion should wait for them accordingly.
+
+If someone who raised an issue does not respond in any way within two weeks of the last
```
ax3l: I admit I sometimes need pretty much three weeks these days, but find two weeks is a very sensible ask 👍 (Note that the average summer vacation can be 2 weeks though, which one can announce as above.)

ax3l (dismissed):
Thanks, agreed! :)

ax3l:
Maybe one small detail we should discuss:
For an open PR, would a maintainer need to refresh their concern every two weeks if the PR author refuses to fix it? :sweat_smile:

source/governance.md
```diff
@@ -109,10 +109,25 @@ Appointed after contributing to PALS (see above) by any member of the TC.
...
+message in that discussion thread, the decision of whether to merge the PR shall proceed
+without further delay. That is, the unanswered issue is treated as resolved and is not
```
ax3l: Maybe to avoid sliding by a Technical Committee member that is on vacation and nobody else taking over:
```suggestion
without further delay, if at least one Technical Committee member agrees. That is, the unanswered issue is treated as resolved and is not
```

DavidSagan: @ax3l If anyone (does not have to be from the Technical Committee) thinks the issue in question is not resolved even after there has been a reply to the original comment, they should post a further comment and this would need to be addressed. I therefore think the wording as is is best.

DavidSagan:
> Maybe one small detail we should discuss: For an open PR, would a maintainer need to refresh their concern every two weeks if the PR author refuses to fix it? 😅

I don't know what is meant by "maintainer" here. In any case, the proposed change does not concern the situation where someone has raised an issue and that issue is not addressed. In this case, the issue needs to be addressed before consideration whether to merge. There are no time limits here. And I put in a clarification on this in the PR.

EZoni:
> > Maybe one small detail we should discuss: For an open PR, would a maintainer need to refresh their concern every two weeks if the PR author refuses to fix it? 😅
> 
> I don't know what is meant by "maintainer" here. In any case, the proposed change does not concern the situation where someone has raised an issue and that issue is not addressed. In this case, the issue needs to be addressed before consideration whether to merge. There are no time limits here. And I put in a clarification on this in the PR.

On the maintainer question, my guess is that it meant the members of the PALS technical committee https://github.com/orgs/pals-project/teams/pals-technical-committee.

EZoni:
> For an open PR, would a maintainer need to refresh their concern every two weeks if the PR author refuses to fix it? 😅

And I think the hidden question here may be, how do two maintainers (the reviewer and the PR author) come to an agreement if they disagree on how to move forward with regard to a specific change? The question is also valid for the case where the PR author is a regular contributor (i.e., not a maintainer), but I think it is more difficult to answer in the first case where both parties are maintainers. In fact there may be no "governance" (i.e., policy) answer to this question.

DavidSagan:
> > For an open PR, would a maintainer need to refresh their concern every two weeks if the PR author refuses to fix it? 😅
> 
> And I think the hidden question here may be, how do two maintainers (the reviewer and the PR author) come to an agreement if they disagree on how to move forward with regard to a specific change? The question is also valid for the case where the PR author is a regular contributor (i.e., not a maintainer), but I think it is more difficult to answer in the first case where both parties are maintainers. In fact there may be no "governance" (i.e., policy) answer to this question.

If there is any type of disagreement then there can be a vote of the technical committee. Is not the existing documentation explicit enough in this regard? In any case, this issue is outside of the scope of this PR.

qianglbl (commented):
@DavidSagan After the merging, can some unaddressed comments been reopened?

DavidSagan:
> @DavidSagan After the merging, can some unaddressed comments been reopened?

There are no restrictions about reopening anything. Reopening issues that have been *settled* is possible. Any reopening needs to be done in a new PR.

qianglbl:
@DavidSagan Thanks. I assume that the new PR can refer to the previous PR.

DavidSagan:
> @DavidSagan Thanks. I assume that the new PR can refer to the previous PR.

Yes. A PR can refer to anything.

source/governance.md
ax3l: That would work. The concern I have on phrasing is that it currently reads like the PR author decides this, and not another reviewer.

Something like this?
```suggestion
without further delay. Other reviewers of the PR can object to this and take over a review thread. Otherwise, the unanswered issue is treated as resolved and is not
```

DavidSagan: I have reworded this paragraph. Please take a look.

DavidSagan:
@ax3l @qianglbl @cemitch99 @EZoni @jlvay Ready for rereview.

EZoni (approved):

EZoni:
@DavidSagan 

Looks good to me now. I let @jlvay, @qianglbl, @cemitch99, and @ax3l chime in further.

cemitch99 (approved):
This looks fine to me.

jlvay (approved):

qianglbl (approved):

ax3l (approved):

ax3l:
All review requests are :heavy_check_mark: , will merge :)

PR #296, opened 2026-08-28
Add Ji to TC.

DavidSagan:
Add Ji to TC.

### DIFF
--- a/source/governance.md
+++ b/source/governance.md
@@ -44,7 +44,6 @@ SC members can resign or be removed by majority vote, e.g., due to inactivity, b
 - Ji Qiang
 - Chad Mitchell
 - Axel Huebl
-- Remi Lehe
 - Edoardo Zoni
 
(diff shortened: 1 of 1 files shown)


cemitch99 (dismissed):

source/governance.md
```diff
@@ -41,6 +41,7 @@ SC members can resign or be removed by majority vote, e.g., due to inactivity, b
 
 - Jean-Luc Vay
 - David Sagan
+- Ji Qiang
```
jlvay, line 44: Looking at the current TC, we should remove Remi Lehe and add Edoard Zoni.

jlvay (changes requested):

qianglbl (approved):

source/governance.md
DavidSagan: Done.

ax3l (commented):
Approved

PR #304, opened 2026-09-16
Remove `expand_lattice` reference since `expand_lattice` no longer exists.

DavidSagan:
Remove `expand_lattice` reference since `expand_lattice` no longer exists.


EZoni:
I think we should remove also the comment here then:

https://github.com/pals-project/pals/blob/08f7939d057938bdf63d5975d07a3a2a0462a247/examples/unit_tests/sets/set_single_definition.pals.yaml#L5

DavidSagan:
@EZoni Done. Ready for rereview.

EZoni (approved):

qianglbl (approved):

ax3l (approved):

PR #305, opened 2026-09-17
Now all angles are radians.

DavidSagan:
Right now the standard is inconsistent with some quantities using radians and others using radians/2pi. The units section specifies angles and phases to be radians / 2pi. This PR converts everything to radians.


EZoni (approved):
I'm fine with this. Not sure if others want to chime in, sometimes people have strong opinions about units.

ax3l (approved):
Thanks, LGTM.

DavidSagan:
> I'm fine with this. Not sure if others want to chime in, sometimes people have strong opinions about units.

I think a discussion is in order as to whether we want to make exceptions in certain cases like RF phases and betatron tunes and this could be something a later PR changes.

qianglbl (approved):

PR #308, opened 2026-09-17
Simulations section removed

DavidSagan:
The simulations section was badly out-of-date and since things are developing quickly will need regular maintenance to keep current. Also it needs to be revamped based on what is being developed by NARAD, etc. My proposal is to remove this section for now and something new can be formulated later.


EZoni:
This PR seems to contain other changes beyond removing the simulations section. Was that intentional or by accident?

DavidSagan:
> This PR seems to contain other changes beyond removing the simulations section. Was that intentional or by accident?

@EZoni Sorry about that. Material from another PR got mixed in. This has been corrected.

source/simulations.md
```diff
@@ -1,12 +0,0 @@
-(c:impl.sims)=
-# Simulations
-
-PALS will be used in various accelerator lattice simulations.
-
-## Planned Implementations
-
-- [BMAD](https://github.com/bmad-sim)
-- [BLAST ImpactX](https://impactx.readthedocs.io)
```
ax3l, line 9: Discussed: ImpactX can already read PALS files :)
https://impactx.readthedocs.io/en/latest/usage/python.html#impactx.elements.KnownElementsList.load_file

ax3l: We can merge this section into `code-packages`

DavidSagan:
Git history is messed up. Closing this and opening a new PR.

PR #311, opened 2026-09-17
New adjustments construct

DavidSagan:
New adjustments construct.


qianglbl (commented):
@DavidSagan 1)line 210, "tune of a machine": I don't quite understand why the tune of a machine will be different from different simulation programs.
2)why the default value of weight for variable is zero?

DavidSagan:
> @DavidSagan 1)line 210, "tune of a machine": I don't quite understand why the tune of a machine will be different from different simulation programs. 2)why the default value of weight for variable is zero?

The tune can be different since different programs use different physics models for a given element and this will lead to differences in the transfer matrix and hence the tune.

The default weight for a variable is zero since there is no other value that would make a sensible default.

qianglbl:
@DavidSagan My understanding of tune is a measure of number of betatron oscillation per turn. For a given periodic lattice, there is only one tune. I don't understand why different programs can generate different tunes? Can you illustrate this a bit more?

DavidSagan:
> @DavidSagan My understanding of tune is a measure of number of betatron oscillation per turn. For a given periodic lattice, there is only one tune. I don't understand why different programs can generate different tunes? Can you illustrate this a bit more?

Correct. Tune is the number of oscillations per turn. For a given particle starting position, the particle position after one turn will be different according to different programs due to differences in the physics models used. These differences will then lead to differences in derived quantities like the tune, Twiss parameters, etc.

qianglbl:
@DavidSagan I see. It looks like you are thinking about the individual particle tune, while I was talking about the global machine tune. Should we make it clear in the document?

DavidSagan:
> @DavidSagan I see. It looks like you are thinking about the individual particle tune, while I was talking about the global machine tune. Should we make it clear in the document?

"Tune", without any qualifications, is always defined as the tune in the limit of zero amplitude of oscillation. So there is nothing to clarify here.

DavidSagan:
Another way to say things is that the 1-turn transfer matrix computed by different programs will be different and this will lead to different tune values.

qianglbl:
@DavidSagan I can understand the 1-turn transfer matrix can be different from different programs depending on the definition of coordinates. However, the number of oscillations should be independent of that definition. Can you give an example that different programs will generate different tunes for a given same lattice?

DavidSagan:
> @DavidSagan I can understand the 1-turn transfer matrix can be different from different programs depending on the definition of coordinates. However, the number of oscillations should be independent of that definition. Can you give an example that different programs will generate different tunes for a given same lattice?

No. For example, Bmad and MAD use the same coordinates but get differing results including transfer matrices. The tune can be calculated directly from the transfer matrix which leads to different tunes as calculated by Bmad and MAD.

qianglbl:
@DavidSagan I don't quite understand why two programs using the same coordinates and the same lattice give different linear one-turn matrix. Is it because different programs have different assumption about the accelerator element model such as fringe field model? When we benchmark two programs, the agreement of the linear one-turn matrix or tune seems to be the first step check. Maybe, I missed something. We can talk about this on Monday's meeting.

DavidSagan:
Yes different programs use different models and if you check any of the comparisons you will see differing results for the tunes and other parameters. Sometimes the comparisons do not print enough digits so the results look the same but in reality the results are never the same.

